Issue #16, opened 2016-11-17
Counter in the network preparation screen is not being updated correctly


pedrocamargo:
![untitled](https://cloud.githubusercontent.com/assets/7377666/20411937/898a3976-ad6f-11e6-9441-1bd6e8112a60.png)

Error is probably when we set the max number of the counter

PR #17, opened 2016-11-17
Fixing the counter for network preparation WITH pre-existing nodes

pedrocamargo:


pedrocamargo:
Fix for the following issue:

Counter in the network preparation screen is not being updated correctly #16

PR #27, opened 2017-02-07
Least common denominator now returns complete sets for both layers an…

pedrocamargo:
…alyzed


pedrocamargo (commented):
Works and should be merged

PR #102, opened 2018-08-12
Pedro/distribution menu

pedrocamargo:


pedrocamargo:
Yu-Chu,

This one includes the data menu as well, so if you review the data one first, this should be a lot smaller.

ychtw (approved):
looks good!

PR #110, opened 2018-10-28
Prevents errors thrown by lack of binaries and adds back download menu

pedrocamargo:
The title is self-explanatory.

I think we will be going towards the release of a preliminary version soon...

### DIFF
--- a/AequilibraEMenu.py
+++ b/AequilibraEMenu.py
@@ -23,8 +23,9 @@
 # noinspection PyUnresolvedReferences
 import os
 import sys
+sys.dont_write_bytecode = True
+import os.path
 import qgis
-import aequilibrae
 # import tempfile, glob
 # from qgis.core import *
 from qgis.PyQt import QtWidgets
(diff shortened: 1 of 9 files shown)


AequilibraEMenu.py
```diff
@@ -69,20 +63,22 @@
 
 # from .public_transport_procedures import GtfsImportDialog
 
+# Procedures that depend on AequilibraE
```
jamiecook, line 67: I always try to avoid exceptions as flow control... If you know what causes the error, can you just check for that directly?

binary_downloader_class.py
```diff
@@ -1,31 +1,30 @@
...
-        QtGui.QDialog.__init__(self)
+        QtWidgets.QDialog.__init__(self)
         self.iface = iface
         self.setupUi(self)
         error = True
         self.linux64 = 'https://goo.gl/V5Zkyv'
         # self.linux32 = 'http://www.aequilibrae.com/binaries/v.0.3.3/AoN_linux64.so'
         self.windows32 = 'https://goo.gl/cA9FDu'
-        self.windows64 = 'https://goo.gl/XY3Ri9'
+        self.windows64 = 'https://github.com/AequilibraE/aequilibrae/releases/download/untagged-e7504e084925a01631c0/AoN.cp36-win_amd64.pyd'
```
jamiecook, line 21: Would be nice if all of these urls were built up from a single commit

ychtw:
I don't have further comment

binary_downloader_class.py
pedrocamargo: They will be...  This is temporary, and will refer to aequilibrae's latests release at the time of the qgis plugin release

AequilibraEMenu.py
```diff
@@ -69,20 +64,20 @@
...
-old_binary = False
-# try:
-#     from aequilibrae.aequilibrae.paths import VERSION_COMPILED as VERSION_GRAPH
-#     if VERSION != VERSION_GRAPH:
-#         old_binary = True
-# except:
-#     no_binary = True
+try:
+    from .aequilibrae.aequilibrae.paths import allOrNothing
+except ImportError:
```
pedrocamargo, line 71: This is what can be done, AFAIK

PR #117, opened 2018-12-30
Pedro/adjustments

pedrocamargo:


pedrocamargo:
This is a REALLY big pull request that corresponds to a pretty thorough check of all tools, with the exception of Public Transportation, which is not included in this release

ychtw (approved):
Looks good!

PR #127, opened 2019-04-03
Pedro/issue 121

pedrocamargo:
Fixes issue 124, which had to do with the progress bar for desire line creation


arobrien:
The description is very similar to title of #125 and has some of the same commits, but not all?

pedrocamargo:
I did one branch on top of the other..

PR #131, opened 2019-04-12
Pedro/issue 128

pedrocamargo:
New, ultra-fast desire lines


pedrocamargo:
It would be useful to merge the other ones first...

PR #142, opened 2019-06-17
fixes issue 138 - Creation of graphs with dual fields

### LINKED ISSUE #138: Creation of graphs with dual fields
Error seems to be that we are not doubling the fields that are actual dual.  Stack trace below

2019-06-17T10:41:52     WARNING    Traceback (most recent call last):
              File "/home/pedro/.local/share/QGIS/QGIS3/profiles/default/python/plugins/AequilibraE/paths_procedures/create_graph_dialog.py", line 291, in finished_threaded_procedure
              block_centroid_flows=self.block_through_centroids)
              File "/home/pedro/.local/share/QGIS/QGIS3/profiles/default/python/plugins/AequilibraE/aequilibrae/aequilibrae/paths/graph.py", line 383, in set_graph
              raise ValueError("cost_field not available in the graph:" + str(self.graph.dtype.names))
             ValueError: cost_field not available in the graph:('link_id', 'a_node', 'b_node', 'direction', 'id', 'length', 'travel_time')

pedrocamargo:
Error only happened when a dual field was set as a the minimizing field

closes #138 

(https://help.github.com/en/articles/closing-issues-using-keywords)

### DIFF
--- a/paths_procedures/create_graph_dialog.py
+++ b/paths_procedures/create_graph_dialog.py
@@ -413,6 +413,9 @@ class GraphCreationDialog(QtWidgets.QDialog, FORM_CLASS):
             if self.cost_field is None:
                 self.error = 'Cost field not selected'
 
+            elif self.cost_field[-2:] == '_*':
+                self.cost_field = self.cost_field[:-2]
+
         if self.error is not None:
             ready = False
 


paths_procedures/create_graph_dialog.py
```diff
@@ -413,6 +413,9 @@ def check_if_ready(self):
             if self.cost_field is None:
                 self.error = 'Cost field not selected'
 
+            elif self.cost_field[-2:] == '_*':
```
jamiecook, line 416: really hard to say what the fixes accomplishes... I read through the issue and the method, not really sure why the cost_field would end with _* or why that should be removed from the string.

If I had any meaningful feedback it would be that the method is too long and does too many things. 

Especially the checking of both error and ready .... I would write a number of smaller methods and wrap them up in something that can early exit when something goes wrong. 

error = check_one()
error = error || check_two()
error = error || check_three()

if error: 
    tell the user
else
    ready = True

jamiecook (approved):

PR #143, opened 2019-08-02
Blacked!

pedrocamargo:
This PR is just intended to start developing some code parameters standards for this repository as well


.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.7
```
ychtw, line 6: QGIS on mac still require python 3.6 now. Will this matter?

common_tools/__init__.py
```diff
@@ -30,4 +30,4 @@
 from .report_dialog import ReportDialog
 from .worker_thread import WorkerThread
 from .about_dialog import AboutDialog
-from .all_layers_from_toc import all_layers_from_toc
\ No newline at end of file
+from .all_layers_from_toc import all_layers_from_toc
```
ychtw: Looks the same as before?

ychtw (commented):
One question about the python version.

.pre-commit-config.yaml
pedrocamargo: No. This is only for running Black itself.

The Python 3.6/3.7 issue is something we will have to navigate, though, as for Windows it is already using Python 3.7.  We might have to look into Continuous Integration at some point soon.

ychtw: Got it. Thanks for the clarification!

ychtw (approved):

PR #145, opened 2019-09-10
Pedro/package installer

pedrocamargo:
Allows the user to install open matrix from inside QGIS, as long as they have started QGIS as administrator

### DIFF
--- a/AequilibraEMenu.py
+++ b/AequilibraEMenu.py
@@ -48,6 +48,7 @@ from .common_tools import ParameterDialog
 from .common_tools import AboutDialog
 
 from .binary_downloader_class import BinaryDownloaderDialog
+from .download_extra_packages_class import DownloadExtraPackages
 from .distribution_procedures import DistributionModelsDialog
 
 from .gis import CompareScenariosDialog
@@ -80,6 +81,12 @@ if not no_binary:
     from .paths_procedures import TrafficAssignmentDialog
     from .paths_procedures import ShortestPathDialog
     from .paths_procedures import ImpedanceMatrixDialog
+
+extra_packages = True
+try:
+    import openmatrix as omx
+except ImportError:
+    extra_packages = False
 
 
 class AequilibraEMenu(object):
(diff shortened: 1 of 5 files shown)


matrix_procedures/display_aequilibrae_formats_dialog.py
```diff
@@ -40,31 +52,51 @@ def __init__(self, iface):
...
         if self.data_type is None:
             self.error = "Path provided is not a valid dataset"
             self.exit_with_error()
-
-        self.data_type = self.data_type.upper()
-
-        if self.data_type == "AED":
-            self.data_to_show = AequilibraEData()
-        elif self.data_type == "AEM":
+        else:
```
jamiecook, line 70: does exit_with_error() not actually exit the method?
The else is superfluous (unless you've got type checking problems)

matrix_procedures/display_aequilibrae_formats_dialog.py
```diff
@@ -40,31 +52,51 @@ def __init__(self, iface):
...
+            self.data_type = self.data_type.upper()
+
+        if self.data_type in ["AED", "AEM"]:
```
jamiecook, line 73:
```suggestion
        if self.data_type in ["AED", "AEM"]:
            load_aequilibrae_formatted_data()
        elif self.data_type == "OMX":
            load_omx_format_data()
```

You can also put in the stuff about computation_views in those methods

AequilibraEMenu.py
```diff
@@ -315,6 +322,14 @@ def initGui(self):
             self.binary_action.triggered.connect(self.run_binary_download)
             self.binary_action.setEnabled(True)
             self.AequilibraE_menu.addAction(self.binary_action)
+
```
jamiecook, line 325: neat

jamiecook (approved):

matrix_procedures/display_aequilibrae_formats_dialog.py
pedrocamargo: It actually goes through for some reason, hence the else

matrix_procedures/display_aequilibrae_formats_dialog.py
pedrocamargo: I need to load the matrix first...

Issue #164, opened 2020-05-03
Create connectors for the OSM network.


zhwlxl:
Hi Pedro, 

Thanks for your great work on Aequilibrae. 

I am trying to make a highway transport network from the openstreet map, but I am stuck at the step of creating the connectors. It did create the connectors but also a new B_node which is overlaping the existing node in the network. Please see the screenshot below:
<img width="766" alt="Screenshot 2020-05-03 at 22 33 18" src="https://user-images.githubusercontent.com/3350470/80926315-7372d780-8d8e-11ea-96c3-99c5a1e31648.png">

Therefore, when I was trying to do skims, it always show some huge numbers, which is no connections between zones. Is that possible you can have a look on this issue, I am not sure if I did something wrong or not. Also, is there a way to speed up the creating connectors process, for about 360 zones, it takes few hours to create the connectors. 

Thanks a lot.

Regards,
Hongwei

pedrocamargo:
Hi @zhwlxl , Cna you share the dataset you are using, as well as the operating system and AequilibraE version you are using?

zhwlxl:
Hi Pedro, thanks for your reply. Please find the link of the project file https://1drv.ms/u/s!ApCAzDcnT4qagfgXaRD9rO2_RdfrTQ?e=lghssj.  It's created from QGIS 3.10.5 on windows 10. I also tried on my Mac, however the QGIS is crached when I load any project file from AE GUI on MacOS, therefore, I copied out the script to run in python which gives the same results as the connector B_node is created overlapping the existing node. here is the script in jupyter format https://1drv.ms/u/s!ApCAzDcnT4qagfgYk_L-PCiiGeSXyQ?e=jq3NEw

Many thanks

pedrocamargo:
@zhwlxl , running it in Python will probably not work due to the lack of the Rtree Sqlite extension on most Python distributions (this is not the Rtree  module), so I will investigate the issue with the plugin first and get back to you

pedrocamargo:
Investigating a little more...

The procedure is NOT creating overlapping nodes, but rather inserting the wrong coordinates for the link extremity that corresponds to the existing network link
 
![image](https://user-images.githubusercontent.com/7377666/81028838-5d683280-8ec6-11ea-9700-74b982ec24d6.png)

[...]

PR #168, opened 2020-05-07
Addresses issue #164

pedrocamargo:


network/adds_connectors_procedure.py
```diff
@@ -50,8 +29,8 @@ def __init__(
         self.max_connectors = max_connectors
         self.allowed_link_types = allowed_link_types
         self.error = None
-        self.insert_qry = """INSERT INTO links (direction, modes, link_type, geometry) 
-                             VALUES({}, GeomFromText("{}", 4326))"""
+        self.insert_qry = '''insert into links (modes, link_type, geometry) values
```
arobrien, line 32: was this supposed to drop setting of 'direction'?

pedrocamargo: Yes.  Direction has a default value of 0 (bidirectional) in the links table.

PR #175, opened 2021-05-10
add processing provider v1

Art-Ev:
First version of processing provider (usable through processing toolbox and QGIS modeler)
Tested on QGIS 3.16 & 3.18 :
image
image


Art-Ev (commented):
image

AequilibraEMenu.py
```diff
@@ -324,14 +327,23 @@ def run_log(self):
 
     def unload(self):
         del self.dock
+        try:
```
pedrocamargo: Broad try/except clauses are not a good workflow, as they tend to be quite broad and will fail Linting and we start using it (probably soon).  Is there a test to check if the provider exists and then remove it?

pedrocamargo (changes requested):
Really nice work.  The only point to improve is that try/except and we can merge it into master!!

processing_provider/Create_Nodes.py
```diff
@@ -0,0 +1,110 @@
...
+            'VERTICES': '0,-1',
+            'OUTPUT': QgsProcessing.TEMPORARY_OUTPUT
+        }
+        outputs['ExtraireLesExtrem'] = processing.run('native:extractspecificvertices', alg_params, context=context, feedback=feedback, is_child_algorithm=True)
+
+        feedback.setCurrentStep(1)
+        if feedback.isCanceled():
+            return {}
+
+        # Supprimer les géométries dupliquées
```
jamiecook:
```suggestion
        # Remove duplicate geometries
```

jamiecook (approved):
Looks good to me, not sure what you want to do about mixing english/french in the comments and intermediate variable names. But overall legit - I think that the QGIS processing engine makes things more verbose than they need to be get-vertices, remove duplicates and filter for non-centroids would be 1-2 lines in pure geopandas for instance... but not much you can do about that if you want to use pure QGIS

AequilibraEMenu.py
Art-Ev: Will take a look at this

Art-Ev:
> Looks good to me, not sure what you want to do about mixing english/french in the comments and intermediate variable names. But overall legit - I think that the QGIS processing engine makes things more verbose than they need to be get-vertices, remove duplicates and filter for non-centroids would be 1-2 lines in pure geopandas for instance... but not much you can do about that if you want to use pure QGIS

I'll be more careful not to mix French and English and I'll make a new pull request in full English. For the use of QGIS processes it's also easier for me for the moment but I'll try with geopandas (maybe some computing time to earn at the same time?)

jamiecook:
Sorry @Art-Ev - I didn't mean those comments directly for your immediate action, it's @pedrocamargo call on the EN/FR thing, and the geopandas comment was more of an observation on the differences between geopandas and QGIS.  I don't think it's possible to use geopandas in QGIS directly so your PR is the right way to do this. If you wanted to modify this PR to use english it would make it consistent with the rest of the codebase - but I'm sure that Pedro has left some Portugese in there somewhere :)

AequilibraEMenu.py
```diff
@@ -327,10 +327,7 @@ def run_log(self):
 
     def unload(self):
         del self.dock
-        try:
-            QgsApplication.processingRegistry().removeProvider(self.provider)
-        except:
-            pass
+        QgsApplication.processingRegistry().removeProvider(self.provider)
```
pedrocamargo: What happens if the provider does not exist?

pedrocamargo: Maybe something like this:

```suggestion
         if  self.provider in QgsApplication.processingRegistry().providers():
             QgsApplication.processingRegistry().removeProvider(self.provider)
```

Art-Ev:
Upload to AWS failed, is it possible to relaunch checks @pedrocamargo ?

PR #191, opened 2022-09-08
Pedro/dev setup docker

pedrocamargo:
@jamiecook , This PR is a first clean and setup. It is intended to include:

- [x] Black the code
- [x] VSCode container setup
- [x] Changes to CI with the indication of the Docker image to use on unit tests


pedrocamargo:
@jamiecook , I have ZERO ideas of why the test is accusing an import from outside the package top level. This works well for QPolaris

pedrocamargo:
The current issue on loading the plugin because of some circular imports inside aequilibrae is proving to be a bigger problem than I had anticipated, so bigger refactoring of AequilibraE where nothing is imported in the __init__ may be necessary

pedrocamargo:
@jamiecook , can you take a look at this error I am getting?

pedrocamargo:
@TTalex , can you take a look if you can fix this PR?

TTalex:
Is this an acceptable solution ?
https://github.com/AequilibraE/AequilibraE-GUI/pull/202

Tests are passing, but I'm not sure if I broke something else (which often happens when moving paths around :/)

pedrocamargo:
Let's merge this one and make a release? @jamiecook @Art-Ev

jamiecook (approved):

PR #202, opened 2022-10-24
Avoid relative paths to make tests pass

TTalex:
Looks like tests are passing without relative paths and  path.inserts in tests/__init__.py file.

Would this be acceptable, or are there side effects in regular project usage ?

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,3 +1,7 @@
 # import qgis libs so that ve set the correct sip api version
 
 import qgis  # pylint: disable=W0611  # NOQA
+import sys
+from os.path import abspath, join, dirname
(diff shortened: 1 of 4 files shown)


jamiecook (approved):

tests/__init__.py
```diff
@@ -1,3 +1,7 @@
 # import qgis libs so that ve set the correct sip api version
 
 import qgis  # pylint: disable=W0611  # NOQA
+import sys
+from os.path import abspath, join, dirname
+sys.path.insert(0, abspath(join(dirname(dirname(__file__)), "")))
```
jamiecook, line 6: probably don't need to join with an empty string :)

Issue #198, opened 2022-10-06
implement some GUI based tests


jamiecook:
Finish Pedro's PR: https://github.com/AequilibraE/AequilibraE-GUI/pull/191

PR #206, opened 2023-01-11
First batch of gui tests based on pytest instead of unittest

TTalex:
Previous tests located in the "tests" folder were based on Qgis plugin builder. However, in latest QGIS versions, the test utility tools provided by the plugin builder are broken because of the use of deprecated objects such as QgsMapLayerRegistry.

I have created a new "test" folder with unit tests based on https://github.com/GispoCoding/pytest-qgis and https://github.com/pytest-dev/pytest-qt

I have added two test files:
* test/test_qaequilibrae_menu_without_project.py : Loads the plugin and tries to click on every menu and checks that the proper dialog is open. Since no project is loaded, most menus trigger an error message, which is also checked.
* test/test_qaequilibrae_menu_with_project.py : Loads the plugin, loads the sioux falls tests project (added in test/data), tries to click on every menu and checks that the proper dialog is open.

Some tests are skipped or incomplete for now:
* Dialogs that open a QFileDialog : I have yet to find a way to capture the FileDialog to interact and close it. I'm not sure if it's possible.
* test_about_menu : This test fails because the "About" button is currently broken (meta files are not found in the modules folder). Here's the traceback for reference:
```
  File "/xx/qaequilibrae/aequilibrae_menu.py", line 278, in run_about
    dlg2 = AboutDialog(self.iface)
  File "/xx/qaequilibrae/modules/common_tools/about_dialog.py", line 33, in __init__
    with open(os.path.join(d, "meta.yaml"), "r") as yml:
FileNotFoundError: [Errno 2] No such file or directory: '/xx/qaequilibrae/modules/meta.yaml'
```
---

To avoid confusions, I deleted the old "tests" folder, and updated the workflows.

Windows are hidden by default using `qgis_qui_enabled=False` in the test/pytest.ini config file.

---

Related issue: https://github.com/AequilibraE/qaequilibrae/issues/198


pedrocamargo:
@TTalex , did you mean to include the path files under tests?

TTalex:
> @TTalex , did you mean to include the path files under tests?

Whoops, I did not, good catch, thanks.

test/data/SiouxFalls_project/aequilibrae.log
```diff
@@ -0,0 +1,4 @@
+2020-08-10 17:44:58,063;aequilibrae;INFO ; Opened project on D:\src\aequilibrae\tests\data/SiouxFalls_project
```
pedrocamargo, line 1: I guess this one can go as well?  Maybe adding to .gitignore as well?

test/test_qaequilibrae_menu_with_project.py
```diff
@@ -0,0 +1,295 @@
...
+    window = QApplication.activeWindow()
+    while window is None and timeout > 0:
+        window = QApplication.activeWindow()
+        qtbot.wait(100)
+        timeout -= 100
+    assert timeout > 0, "Waiting for window to open timed out after 3 seconds"
+    return window
+
+
+def check_if_new_active_window_matches_class(qtbot, windowClass):
```
pedrocamargo, line 19: Nice!

test/test_shortest_path.py
```diff
@@ -0,0 +1,91 @@
+import os
+import pytest
+from PyQt5.QtCore import Qt, QTimer
+from PyQt5.QtWidgets import QApplication
+from qgis.core import QgsProject, Qgis, QgsVectorLayer
+from qaequilibrae.modules.paths_procedures.show_shortest_path_dialog import ShortestPathDialog
+
+
+def load_layers():
```
pedrocamargo, line 9: Very nice!!!

pedrocamargo (approved):

pedrocamargo:
@TTalex I am sorry for the delay in reviewing.

test/data/SiouxFalls_project/aequilibrae.log
TTalex: For sure, I'll add it to the .gitignore.
Btw, we could also drop it from the python repo: https://github.com/AequilibraE/aequilibrae/blob/develop/tests/data/SiouxFalls_project/aequilibrae.log

TTalex:
> @TTalex I am sorry for the delay in reviewing.

All good, thanks, I'm glad this is useful !

PR #208, opened 2023-04-25
196 switch installation of extra packages to users folder

### LINKED ISSUE #196: Switch installation of extra packages to user's folder
Installing extra packages for OMX support and Routing has proven to be an issue for many users, as that requires administrator privileges, and doing it without often breaks QGIS and/or AequilibraE.

Looking at a user's path (on Windows), we see the following. 

**C:\Users\USER\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\AequilibraE-GUI\aequilibrae**
C:/PROGRA~1/QGIS32~2.1/apps/qgis/./python
C:/Users/USER/AppData/Roaming/QGIS/QGIS3\profiles\default/python
C:/Users/USER/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins
C:/PROGRA~1/QGIS32~2.1/apps/qgis/./python/plugins
C:\Program Files\QGIS 3.26.1\bin\python39.zip
C:\PROGRA~1\QGIS32~2.1\apps\Python39\DLLs
C:\PROGRA~1\QGIS32~2.1\apps\Python39\lib
C:\Program Files\QGIS 3.26.1\bin
**C:\Users\USER\AppData\Roaming\Python\Python39\site-packages**
C:\PROGRA~1\QGIS32~2.1\apps\Python39
C:\PROGRA~1\QGIS32~2.1\apps\Python39\lib\site-packages
C:\PROGRA~1\QGIS32~2.1\apps\Python39\lib\site-packages\win32
C:\PROGRA~1\QGIS32~2.1\apps\Python39\lib\site-packages\win32\lib
C:\PROGRA~1\QGIS32~2.1\apps\Python39\lib\site-packages\Pythonwin
C:/Users/USER/AppData/Roaming/QGIS/QGIS3\profiles\default/python
C:\Users/USER/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\go2streetview
C:\Users\USER\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\go2streetview\res

Can't we install those packages in one of the paths highlighted in bold (or in the AequilibraE-GUI folder under a subdirectory)?
Local tests suggest we can, so applying the change would be terrific.

In case we can, it would be ideal to have the AequilibraE library changed to an external package as well, which would automatically install the required version of the library without requiring the current janky workflow with the binaries.

The one issue that might make it impossible to make the AequilibraE Python package not installable through PIP is that NumPy changes the size of its C Array, which breaks compatibility for compiled code such as that in AequilibraE. Maybe releasing always 2+ versions of AequilibraE (Python) for different versions of C-arrays in NumPy?

pedrocamargo:


download_extra_packages_class.py
```diff
@@ -2,31 +2,30 @@
...
-        with open(os.path.dirname(__file__) + "/requirements.txt", "r") as req:
-            self.packages = [line.rstrip() for line in req.readlines()]
+        pth = os.path.dirname(__file__)
+        self.file = join(pth, "requirements.txt")
+        self.pth = join(pth, "packages")
 
     def install(self):
+        spec = iutil.find_spec("aequilibrae")
+        if spec is not None:
+            return
```
jamiecook, line 18: Is this skipping if Aeq is installed? what happens if we add things to the requirements.txt and re-run?

modules/public_transport_procedures/gtfs_import_dialog.py
```diff
@@ -1,7 +1,7 @@
 import logging
 import os
 
-from aequilibrae.transit.gtfs import create_gtfsdb
+# from aequilibrae.transit.gtfs import create_gtfsdb
```
jamiecook, line 4: ?

requirements.txt
```diff
@@ -1,3 +1,3 @@
-aequilibrae
+aequilibrae==0.8.4.dev1
```
jamiecook, line 1: are you planning to lock it to a specific version?

jamiecook (approved):

jamiecook (approved):

requirements.txt
pedrocamargo: Yes. That's how we get it to install a version that the same version of NumPy as QGIS
image

download_extra_packages_class.py
pedrocamargo: The idea is that the plugin will install all the requirements it has, and after that it is done. In development mode you need to do a little more work...

modules/public_transport_procedures/gtfs_import_dialog.py
pedrocamargo: We still have code there for the old PT importer.  I chose to keep if there and connect it to the new one a little later.

PR #213, opened 2023-05-10
Try to use the right python to install packages

jamiecook:
This was causing problems as the version of python used to install packages might not be the same version as the python QGIS actually uses to run our plugin.

There still seems to be a problem with the linux approach. Jake is having difficulty.

### DIFF
--- a/download_extra_packages_class.py
+++ b/download_extra_packages_class.py
@@ -16,7 +16,7 @@ class download_all:
 
     def install(self):
         lines = []
-        command = f'{self.find_python()} -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
+        command = f'"{self.find_python()}" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
         print(sys.executable)
         print(command)
         lines.append(command)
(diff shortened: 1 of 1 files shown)


download_extra_packages_class.py
```diff
@@ -15,19 +16,44 @@ def __init__(self):
...
+        # On mac/windows sys.exeuctable returns '/Applications/QGIS.app/Contents/MacOS/QGIS' or
+        # 'C:\\Program Files\\QGIS 3.30.0\\bin\\qgis-bin.exe' respectively so we need to explore in that area
+        # of the filesystem
+        elif sys.platform == "darwin":
+            python_exe = sys_exe.parent / "bin" / "python3"
+        elif sys.platform == "win32":
+            python_exe = sys_exe.parent / "python3"
+
+        if not python_exe.exists():
+            raise FileExistsError("Can't find a python executable to use")
```
pedrocamargo, line 57: I would just return python in this case (to continue the current behaviour. At least there is a chance it will work...

pedrocamargo (approved):

PR #225, opened 2023-06-19
Fixes matrix importer

### LINKED ISSUE #222: Import Matrix
**Describe the bug**
The window to import a matrix using 'Open layer' is showing the option 'Open' instead 'Save'', so it is not saving the .aem file.
There are two images to illustrate:
image
image
 

**To Reproduce**
0. GIS
1. Go to 'Desire Lines'
2. There is no option to LOAD MATRIX

Other option to Load Matrix
0. 'Data'
1. 'Import matrices'
2. Go to 'Open layer' option
3. Set the values and click 'LOAD'
4. Click on 'Save'
5. Open a window with 'Open' instead 'Save'
6. There is no possible to save the matrix

**Expected behavior**
Expected to load or save the matrix.

 - OS: Windows 10
 - QGIS Version 3.22.14

**Additional context**
The 'Load matrices' window open in 'Trip Distribution' and doesn't wort as well.

r-akemii:


test/test_load_matrix.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def test_save_matrix(ae_with_project, qtbot):
+    load_layers()
+    dialog = LoadMatrixDialog(ae_with_project)
+    dialog.show()
+
+    qtbot.mouseClick(dialog.but_load, Qt.LeftButton)
+    qtbot.mouseClick(dialog.but_permanent_save, Qt.LeftButton)
+    dialog.close()
```
pedrocamargo, line 75: We need to open the resulting matrix and check if the data is there (total for the matrix and indices themselves.

pedrocamargo:
Renaming the matrix inside the matrices table in the database is something that has to be investigated as a possible issue

test/test_load_matrix.py
pedrocamargo: @r-akemii , You are NOT testing the output here. If anything, you have shown that running the software results into an empty matrix, which is NOT what should've happen.

pedrocamargo (approved):

PR #232, opened 2023-07-11
fixes null assignment

r-akemii:
Adds *test_traffic_assignment.py*


test/test_traffic_assignment.py
```diff
@@ -0,0 +1,63 @@
...
+    pth = Path("test/data/SiouxFalls_project")
+    results = pth / "results_database.sqlite"
+    assert isfile(results)
+    con = sqlite3.connect(results)
+    assert con.execute(f"SELECT SUM(PCE_tot) FROM {test_name}").fetchone()[0] > 881_000
+
+    num_cores = dialog.assignment.cores
+    log_ = pth / "aequilibrae.log"
+    assert isfile(log_)
+
```
pedrocamargo, line 77: Add test of skimming

test/test_traffic_assignment.py
```diff
@@ -0,0 +1,63 @@
+from pathlib import Path
+from os.path import isfile
+import sqlite3
+from uuid import uuid4
+from PyQt5.QtCore import QTimer, Qt
+from qaequilibrae.modules.matrix_procedures.load_project_data import LoadProjectDataDialog
+from qaequilibrae.modules.paths_procedures.traffic_assignment_dialog import TrafficAssignmentDialog
+
+
+def test_ta_menu(ae_with_project, qtbot):
```
pedrocamargo, line 14: Let's test with two classes, and one of them with more than one matrix core

qaequilibrae/modules/paths_procedures/traffic_assignment_dialog.py
```diff
@@ -28,10 +28,11 @@
 
 
 class TrafficAssignmentDialog(QtWidgets.QDialog, FORM_CLASS):
-    def __init__(self, qgis_project):
+    def __init__(self, qgis_project, testing=False):
```
jamiecook: i don't think it's a good idea to allow our test suite to directly affect the behaviour of our library code.

I would prefer to see an argument 'matrix_view' that changes just the behaviour we want to be different in testing. 

Not sure what to do about the early return though

r-akemii: Hey Jamie! I removed this testing=False and the related pieces of code from the trafic_assignment_dialog and made some changes in the traffic assignment test.

pedrocamargo (approved):

PR #240, opened 2023-07-24
Fixes plugin internationalization

r-akemii:


r-akemii:
@pedrocamargo I believe this PR should solve the past GH Actions failures. I have rewritten the translation and packaging workflow to open a PR when a modification in the translatable files occurs. From what I've seen in the Actions tab, the errors happened because there were modifications in the translatable files, which can only be added to the develop branch (and thus synced with Transifex) through a PR.

pedrocamargo (approved):

PR #242, opened 2023-08-17
Fix automatic package installation when using conda forge qgis

m-richards:
I've been experimenting with using QGIS installed via conda-forge (same idea as [this](https://anitagraser.com/2023/01/21/pyqgis-jupyter-notebooks-on-windows-using-conda/)) as a way to keep qgis up to date more easily, and I noticed that the automatic pip install as part of the plugin download is broken when qgis is packaged in this way.

This was the traceback I got
```python
  File "C:\Users/Matt/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\qaequilibrae.py", line 33, in 
    from aequilibrae.project import Project
  File "C:\Users/Matt/.conda/envs/qgis/Library/./python\qgis\utils.py", line 888, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
ModuleNotFoundError: No module named 'aequilibrae'

During handling of the above exception, another exception occurred:

  File "C:\Users/Matt/.conda/envs/qgis/Library/./python\qgis\utils.py", line 423, in _startPlugin
    plugins[packageName] = package.classFactory(iface)
  File "C:\Users/Matt/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\__init__.py", line 3, in classFactory
    from .qaequilibrae import AequilibraEMenu
  File "C:\Users/Matt/.conda/envs/qgis/Library/./python\qgis\utils.py", line 888, in _import
    mod = _builtin_import(name, globals, locals, fromlist, level)
  File "C:\Users/Matt/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\qaequilibrae.py", line 43, in 
    result = download_all().install()
  File "C:\Users/Matt/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\download_extra_packages_class.py", line 25, in install
    command = f'"{self.find_python()}" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
  File "C:\Users/Matt/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\download_extra_packages_class.py", line 64, in find_python
    raise FileExistsError("Can't find a python executable to use")
FileExistsError: Can't find a python executable to use
```
and it seemed pretty straightforward to just add another condition to the windows path lookup.

As an aside, if the default was to use plain python `python`, rather than crash with a FileExistsError (or at least attempt to do so), as it looks like the behaviour was before #213 this would have worked without issue for me (based on testing, that, I wasn't able to install and old commit directly to check).

If this is considered out of scope that's fine too, I don't know if the assumption that qgis is bundled with a specific version of python is made, but installing from conda forge would break that - as you can specify qgis and python versions explicitly. 

(I also added some things to the gitignore that showed, but that might be a side effect of how I did my dev setup)

### DIFF
--- a/qaequilibrae/download_extra_packages_class.py
+++ b/qaequilibrae/download_extra_packages_class.py
@@ -58,15 +58,7 @@ class download_all:
         elif sys.platform == "darwin":
             python_exe = sys_exe.parent / "bin" / "python3"
         elif sys.platform == "win32":
-            python_exe = sys_exe.parent / "python3.exe"
-            if python_exe.exists():
-                return python_exe    
-            # Conda sys_exe is ~/.conda/envs/<envname>/Library/bin/qgis.exe,
-            # python is at ~/.conda/envs/<envname>/python.exe
-            conda_candidate = (sys_exe.parent.parent.parent / "python.exe")
-            if conda_candidate.exists():
-                return conda_candidate
-
+            python_exe = Path(sys.base_prefix) / "python.exe"
 
         if not python_exe.exists():
             raise FileExistsError("Can't find a python executable to use")
--- a/qaequilibrae/download_extra_packages_class.py
+++ b/qaequilibrae/download_extra_packages_class.py
@@ -59,9 +59,17 @@ class download_all:
             python_exe = sys_exe.parent / "bin" / "python3"
         elif sys.platform == "win32":
             python_exe = sys_exe.parent / "python3.exe"
+            if python_exe.exists():
+                return python_exe    
+            # Conda sys_exe is ~/.conda/envs/<envname>/Library/bin/qgis.exe,
+            # python is at ~/.conda/envs/<envname>/python.exe
+            conda_candidate = (sys_exe.parent.parent.parent / "python.exe")
+            if conda_candidate.exists():
+                return conda_candidate
+
 
         if not python_exe.exists():
-            raise FileExistsError("Can't find a python executable to use")
+            raise FileNotFoundError(f"Can't find a python executable to use")
         return python_exe
 
     def adapt_aeq_version(self):
--- a/qaequilibrae/download_extra_packages_class.py
+++ b/qaequilibrae/download_extra_packages_class.py
@@ -22,7 +22,7 @@ class download_all:
         self.adapt_aeq_version()
 
         lines = []
-        command = f'"{self.find_python()}" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
+        command = f'"python" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
         print(sys.executable)
         print(command)
         lines.append(command)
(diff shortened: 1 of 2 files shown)


qaequilibrae/download_extra_packages_class.py
```diff
@@ -22,7 +22,7 @@ def install(self):
         self.adapt_aeq_version()
 
         lines = []
-        command = f'"{self.find_python()}" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
+        command = f'"python" -m pip install -r "{self.file}" -t "{self.pth}" --upgrade'
```
pedrocamargo: You removed the search for the python executable to use, so all the logic below becomes useless?

pedrocamargo:
@m-richards , nice to see you around here!!  QGIS does ship with a specific version of Python on Windows, so that is definitely necessary for the average user

qaequilibrae/download_extra_packages_class.py
m-richards: yeah oops, this was testing, not supposed to be commited.

m-richards:
> @m-richards , nice to see you around here!! QGIS does ship with a specific version of Python on Windows, so that is definitely necessary for the average user

Nice to finally have a small excuse to to play with some of aequilibrae. 

Yep, I'm aware that's the usual situation but I don't know if violating that assumption would break the plugin. I see aequilibrae has wheels per platform and python minor release, so as long as the right python is detected, there shouldn't be any issues? I suppose the conda case could be further boxed off by checking for "conda" in sys.version but this seems flakey.

I notice the CI is having some trouble, It looks like one is not looking for a branch on a fork, and the other is looking for secrets which I guess also wouldn't be on a fork.

qaequilibrae/download_extra_packages_class.py
```diff
@@ -59,6 +59,14 @@ def find_python(self):
...
         elif sys.platform == "win32":
             python_exe = sys_exe.parent / "python3.exe"
+            if python_exe.exists():
+                return python_exe    
+            # Conda sys_exe is ~/.conda/envs/<envname>/Library/bin/qgis.exe,
+            # python is at ~/.conda/envs/<envname>/python.exe
+            conda_candidate = (sys_exe.parent.parent.parent / "python.exe")
+            if conda_candidate.exists():
+                return conda_candidate
+
```
m-richards: I noticed the current approach as of #213 also fails for OSGeo4W qgis installs. Might be missing something, but I think all this special casing could just be replaced with this on windows (perhaps other platforms too?)
```suggestion
            python_exe = Path(sys.base_prefix) / "python.exe"
```
As sys.base_prefix seems to be accurate, unlike sys.executable

jamiecook:
Hey @m-richards - sorry for delay in reviewing - i just tested out your suggestion to use sys.base_prefix and can confirm it works nicely for a stock windows install. If it also works for conda i think we'd take a punt that it's going to work for OSGEO installations too.

jamiecook (approved):

PR #250, opened 2023-10-09
Updated processing provider with QGIS processing & better QGIS experience

Art-Ev:
Updated version of processing provider in #175 with a simple export from omx file to a set of csv files

### DIFF
--- a/qaequilibrae/i18n/qaequilibrae_fr.ts
+++ b/qaequilibrae/i18n/qaequilibrae_fr.ts
@@ -48,44 +48,34 @@
 <context>
     <name>AequilibraEMenu</name>
     <message>
-        <location filename="../qaequilibrae.py" line="122"/>
+        <location filename="../qaequilibrae.py" line="118"/>
         <source>Project</source>
         <translation>Projet</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="130"/>
-        <source>Network Manipulation</source>
-        <translation>Manipulation du réseau</translation>
-    </message>
-    <message>
-        <location filename="../qaequilibrae.py" line="176"/>
+        <location filename="../qaequilibrae.py" line="136"/>
         <source>Data</source>
         <translation>Données</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="141"/>
+        <location filename="../qaequilibrae.py" line="140"/>
         <source>Trip Distribution</source>
         <translation>Distribution des déplacements</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="154"/>
+        <location filename="../qaequilibrae.py" line="152"/>
         <source>Paths and assignment</source>
         <translation>Chemins et affectation</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="160"/>
+        <location filename="../qaequilibrae.py" line="158"/>
         <source>Routing</source>
         <translation>Planification d&apos;itinéraire</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="164"/>
+        <location filename="../qaequilibrae.py" line="163"/>
         <source>Public Transport</source>
         <translation>Transports en commun</translation>
-    </message>
-    <message>
-        <location filename="../qaequilibrae.py" line="180"/>
-        <source>Utils</source>
-        <translation>Utilitaires</translation>
     </message>
     <message>
         <location filename="../qaequilibrae.py" line="114"/>
@@ -93,32 +83,32 @@
         <translation>Ouvrir un projet</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="115"/>
+        <location filename="../qaequilibrae.py" line="121"/>
         <source>Create project from OSM</source>
         <translation>Créer un projet à partir d&apos;OSM</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="116"/>
+        <location filename="../qaequilibrae.py" line="124"/>
         <source>Create Project from layers</source>
         <translation>Créer projet à partir de couches</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="119"/>
+        <location filename="../qaequilibrae.py" line="131"/>
         <source>Add zoning data</source>
         <translation>Ajouter des données au zonage</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="120"/>
+        <location filename="../qaequilibrae.py" line="116"/>
         <source>Parameters</source>
         <translation>Paramètres</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="121"/>
+        <location filename="../qaequilibrae.py" line="117"/>
         <source>logfile</source>
         <translation>fichier de log</translation>
     </message>
     <message>
-        <location filename="../qaequilibrae.py" line="122"/>
+        <location filename="../qaequilibrae.py" line="118"/>
         <source>Close project</source>
         <translation>Fermer le projet</translation>
     </message>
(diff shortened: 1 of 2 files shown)


pedrocamargo:
@Art-Ev , I gave this one a test and have a few comments/questions:

1. The advanced section of the procedure has parameters for geo-operations, which I assume are part of some template. Can we remove them?
2. The output of the algorithm is a matrix without indices (zone numbers), when I would expect to have something like "origin, destination, volume", and all in the same table.  Is the intention really this?
3. We need documentation for this section. Happy to help you with that if you'd like
4. The translation scheme was not used, so this would not show in other languages

Art-Ev:
@pedrocamargo : 
1. I don't understand which part of the code you're talking about, in which file?
2. Was quicker (laziness...) and was thinking that it's the same format as EMME csv matrices. After checking (and some searches on openmatrix github) : have updated to include indices and better number formating
3. Where can I start to update documentation ? https://github.com/AequilibraE/qaequilibrae/tree/develop/docs/source ? (do you have a good "model" for me to use as reference ?)
4. How can I declare string variables for translation ? With self.tr('my_string') ?

Art-Ev:
AequilibraE processings are usable by QGIS graphical modeller for "no-code" :
AequilibraE_modeler_Example

pedrocamargo:
@Art-Ev 

1. When you go to advanced-algorithm settings

image

It is also returning an error after the latest change. I fixed the bug, there is a conceptual issue that I am commenting on at the appropriate line of code. Testing would be good...

2. Nice


3. Yes. The softwaredevelopment.rst file is a good one

4. https://aequilibrae.com/qgis/latest/development/softwaredevelopment.html#translatable-strings

qaequilibrae/modules/processing_provider/omx2csv.py
```diff
@@ -0,0 +1,89 @@
...
+            
+            feedback.setCurrentStep(0)
+            feedback.pushInfo('')
+            feedback.pushInfo('Dimensions of matrices to process: ')
+            feedback.pushInfo(str(matrix.shape()))
+            
+            feedback.pushInfo('')
+            feedback.pushInfo('Attributes in omx file: ')
+            feedback.pushInfo(str(matrix.list_all_attributes()))
+            Destination=np.array([list(matrix.mapping('main_index'))])
```
pedrocamargo: Indices have arbitrary names, not main_index only.

File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\openmatrix\File.py", line 214, in mapping
entries.extend(self.get_node(self.root.lookup, title)[:])
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\tables\file.py", line 1600, in get_node
node = where._v_file._get_node(nodepath)
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\tables\file.py", line 1556, in _get_node
node = self._node_manager.get_node(nodepath)
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\tables\file.py", line 417, in get_node
node = self.node_factory(key)
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\tables\group.py", line 1137, in _g_load_child
node_type = self._g_check_has_child(childname)
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\tables\group.py", line 375, in _g_check_has_child
raise NoSuchNodeError(
tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``/lookup/main_index``

During handling of the above exception, another exception occurred:

File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\modules\processing_provider\omx2csv.py", line 42, in processAlgorithm
Destination=np.array([list(matrix.mapping('main_index'))])
File "C:\Users/pcamargo/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qaequilibrae\packages\openmatrix\File.py", line 224, in mapping
raise LookupError('No such mapping: '+title)
LookupError: No such mapping: main_index

Execution failed after 0.03 seconds

qaequilibrae/modules/processing_provider/omx2csv.py
```diff
@@ -0,0 +1,89 @@
...
+    def processAlgorithm(self, parameters, context, model_feedback):
+        results = {}
+        outputs = {}
+        
+        pathSource=parameters['omxFile']
+        pathDest=parameters['destFolder']
+        
+        if has_omx:
+            import openmatrix as omx
+            matrix=omx.open_file(pathSource)
```
pedrocamargo: Why not use the built-in algorithm for exporting matrices?

mat.export('/tmp/my_new_path.csv')

Art-Ev:
**Changes summary :** 
- Translation should be possible now
- had to use an aem temp file with "create_from_omx" function for omx export (was getting the "this make no sense for an omx file" error)
- advanced-algorithm settings : are coming from QGIS and cannot be disable (are not even mentioned in the scripts, coming directly from QGIS processing)

**Still to do** (maybe after a first validation) : update documentation

**For later** : optimize 'create project from link layer' & 'Create/renumber nodes from centroid layer' as sqlite database is written one link/node at a time which result in a pretty slow processing time

Art-Ev:
After some testing, the optimization done on the ProjectFromLayer algorithm reduces processing times by a factor of 10 (ok, 9.5) and allow for a small (not quantified) gain in RAM usage. Will definitively have to port the same method for "centroids update/creation" and eventually "connector creation" (in this case it means doing without AequilibraE python API to use QGIS API and letting the sql triggers do the rest of the magic). 
Process is still using only one core which still leaves a huge potential for optimisation but I think this is some restriction coming from spatialite

qaequilibrae/modules/processing_provider/MatrixFromLayer.py
```diff
@@ -0,0 +1,135 @@
...
+                    QgsProcessingParameterString('CoreName', self.tr('Name for the core of your matrix'), multiLine=False, defaultValue='Value')
+                    ]
+        for param in advparams:
+            param.setFlags(param.flags() | QgsProcessingParameterDefinition.FlagAdvanced)
+            self.addParameter(param)
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        feedback = QgsProcessingMultiStepFeedback(3, model_feedback)
+        results = {}
+        outputs = {}
```
jamiecook: copy pasted variables which aren't used.

qaequilibrae/modules/processing_provider/AssignFromYaml.py
```diff
@@ -0,0 +1,169 @@
...
+class TrafficAssignYAML(TranslatableAlgorithm):
+
+    def initAlgorithm(self, config=None):
+        self.addParameter(QgsProcessingParameterFile('confFile', self.tr('Configuration file (.yaml)'), behavior=QgsProcessingParameterFile.File, fileFilter=self.tr('Assignment configuration file (*.yaml)'), defaultValue=None))
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        feedback = QgsProcessingMultiStepFeedback(5, model_feedback)
+        results = {}
+        outputs = {}
+
```
jamiecook: copy pasted variables which aren't used.

qaequilibrae/modules/processing_provider/Add_connectors.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def initAlgorithm(self, config=None):
+        self.addParameter(QgsProcessingParameterNumber('nb_conn', self.tr('Desired number of connectors'), type=QgsProcessingParameterNumber.Integer, minValue=1, maxValue=10, defaultValue=1))
+        self.addParameter(QgsProcessingParameterString('mode', self.tr('Mode to connect (only one at a time)'), multiLine=False, defaultValue='c'))
+        self.addParameter(QgsProcessingParameterFile('PrjtPath', self.tr('AequilibraE project'), behavior=QgsProcessingParameterFile.Folder, defaultValue='D:/'))
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        feedback = QgsProcessingMultiStepFeedback(2, model_feedback)
+        results = {}
+        outputs = {}
```
jamiecook: copy pasted variables which aren't used.

qaequilibrae/modules/processing_provider/Add_connectors.py
```diff
@@ -0,0 +1,70 @@
...
+        
+        # Checks if we have access to aequilibrae library
+        has_aeq = iutil.find_spec("aequilibrae") is not None
+        if not has_aeq:
+            sys.exit(tr('AequilibraE library not found'))
```
jamiecook:
```suggestion
        if iutil.find_spec("aequilibrae") is None:
            sys.exit(tr('AequilibraE library not found'))
```

qaequilibrae/modules/processing_provider/AssignFromYaml.py
```diff
@@ -0,0 +1,169 @@
...
+        # Opening project
+        project = Project()
+        project.open(params["Project"])
+
+        # Creating graph
+        project.network.build_graphs()
+
+        # Creating traffic classes
+        traffic_classes=[]
+        feedback.pushInfo(str(len(params["Traffic_classes"]))+tr(' traffic classes have been find in config file :'))
```
jamiecook:
```suggestion
        feedback.pushInfo(str(len(params["Traffic_classes"]))+tr(' traffic classes have been found in config file :'))
```

qaequilibrae/modules/processing_provider/AssignFromYaml.py
```diff
@@ -0,0 +1,169 @@
...
+                demand.computational_view([classes[traffic]["matrix_core"]])
+
+                # Getting graph
+                graph = project.network.graphs[classes[traffic]["network_mode"]]
+                graph.set_graph("travel_time")
+                graph.set_blocked_centroid_flows(False)
+
+                if "skims" in classes[traffic] and classes[traffic]["skims"] is not None:
+                    skims=[]
+                    for sk in classes[traffic]["skims"].split(","): skims.append(sk.strip())
```
jamiecook:
```suggestion
                    skims=[sk.strip() for sk in classes[traffic]["skims"].split(",")]
```

qaequilibrae/modules/processing_provider/MatrixFromLayer.py
```diff
@@ -0,0 +1,135 @@
...
+        mat = AequilibraeMatrix()
+        mat.name=matrix_name
+        mat.description=matrix_description
+
+        kwargs = {'file_name': filename,
+                  'zones': num_zones,
+                  'matrix_names': core_name,
+                  'memory_only': False}
+
+        mat.create_empty(**kwargs)
```
jamiecook: Was there a reason to not just use keywords directly?

```suggestion
        mat.create_empty(file_name = filename,
                         zones = num_zones,
                         matrix_names = core_name,
                         memory_only = False)
```

qaequilibrae/modules/processing_provider/RenumberFromCentroids.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def initAlgorithm(self, config=None):
+        self.addParameter(QgsProcessingParameterVectorLayer('nodes', self.tr('Centroids'), types=[QgsProcessing.TypeVectorPoint], defaultValue='nodes'))
+        self.addParameter(QgsProcessingParameterField('node_id', self.tr('Zone id'), type=QgsProcessingParameterField.Numeric, parentLayerParameterName='nodes', allowMultiple=False, defaultValue='node_id'))
+        self.addParameter(QgsProcessingParameterFile('PrjtPath', self.tr('AequilibraE project'), behavior=QgsProcessingParameterFile.Folder, defaultValue='D:/'))
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        feedback = QgsProcessingMultiStepFeedback(3, model_feedback)
+        results = {}
+        outputs = {}
```
jamiecook: again

qaequilibrae/modules/processing_provider/RenumberFromCentroids.py
```diff
@@ -0,0 +1,129 @@
...
+        
+        #Import QGIS layer as a panda dataframe
+        layer = self.parameterAsVectorLayer(parameters, 'nodes', context)
+        columns = [f.name() for f in layer.fields()] + ['geometry']
+        columns_types = [f.typeName() for f in layer.fields()] #
+        row_list = []
+        for f in layer.getFeatures():
+            geom=f.geometry()
+            geom.transform(QgsCoordinateTransform(layer_crs, aeq_crs, QgsProject.instance()))
+            row_list.append(dict(zip(columns, f.attributes() + [geom.asWkt()])))
```
jamiecook: I'll often just use a list comp + helper function for things like this

```suggestion
        def fn(f):
            geom=f.geometry()
            geom.transform(QgsCoordinateTransform(layer_crs, aeq_crs, QgsProject.instance()))
            return dict(zip(columns, f.attributes() + [geom.asWkt()]))
        row_list = [fn(f) for f in layer.getFeatures()]
```

qaequilibrae/modules/processing_provider/exportMatrix.py
```diff
@@ -0,0 +1,64 @@
...
+class exportMatrix(TranslatableAlgorithm):
+
+    def initAlgorithm(self, config=None):
+        self.addParameter(QgsProcessingParameterFile('srcFile', self.tr('Matrix file to convert (.omx or .aem)'), behavior=QgsProcessingParameterFile.File, fileFilter=self.tr('AequilibraE matrix formats (*.omx *.aem)'), defaultValue=None))
+        self.addParameter(QgsProcessingParameterFile('destFolder', self.tr('Export matrices to'), behavior=QgsProcessingParameterFile.Folder, defaultValue=None))
+        self.addParameter(QgsProcessingParameterEnum('outputformat', self.tr('Format to use for export'), options=['.csv','.omx','.aem'], defaultValue='.csv'))
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        results = {}
+        outputs = {}
```
jamiecook: again

qaequilibrae/modules/processing_provider/exportMatrix.py
```diff
@@ -0,0 +1,64 @@
...
+
+        pathSource=parameters['srcFile']
+        fileformat=['.csv','.omx','.aem']
+        pathDest=os.path.join(parameters['destFolder'], Path(pathSource).stem+fileformat[parameters['outputformat']])
+
+        # Checks if we have access to aequilibrae library
+        has_aeq = iutil.find_spec("aequilibrae") is not None
+
+        if not has_aeq:
+            sys.exit(self.tr('AequilibraE library not found'))
```
jamiecook: can we extract this into a helper method?
```suggestion

        def ensure_aeq_available():
            if iutil.find_spec("aequilibrae") is None:
                sys.exit(self.tr('AequilibraE library not found'))
```

and then just call that method in all these providers

Art-Ev: Good idea @jamiecook, will do that once we get a solution for #256 which prevents me from doing any more checking at the moment

Art-Ev:
@pedrocamargo, package download should be good now and translation should also work for directly displayed menus (if .qm file is added by compiling .ts file) : 
- Needed to use a global "fr" language instead of "fr_FR", we may have to change Transifex parameter to output fr translation instead of "fr_FR" (I also think it'll be easier to manage only 1 french translation)
image

- Have created the #256 issue for the infinite qgis opening loop. The qAequilibraE version I used for yesterday training (which include processing provider) was working but not this one 😢

PR #260, opened 2023-12-08
Loads vector from layer

### LINKED ISSUE #255: Vector Loader for IPF is not working
**Describe the bug**
The function to load vectors on Trip Distribution -> IPF is not working. We are able to click on the buttons, but neither *Use data* nor *Save and Use* are loading the data in the vector tab. 

**To Reproduce**

0. Load an AequilibraE Project
1. Load .csv or any other external file containing data
2. Do Trip Distribution -> Iterative Proportional Fitting -> Load 
3. Select the file and set the index field
4. Click one of *Use data* or *Save and Use*

**Expected behavior**
Data would be loaded and appear in the Vector tab.

**Screenshots**
image

 - OS: Windows 11
 - QGIS Version: 3.28.10

r-akemii:
Solves #255. Allows the usage of a vector from a layer in IPF or Apply Gravity procedures. 

The vector name displayed is going to be random, generated by _aequilibraeData_.


pedrocamargo (changes requested):
No tests related to this?

r-akemii:
Not yet. I was expecting to add this test with the ones I created in #254 when I found out about this issue.

PR #263, opened 2023-12-11
Sets tables version in requirements.txt

### LINKED ISSUE #256: qaequilibrae is broken and create infinite QGIS opening loops
#240 seems to have broken qaequilibrae.

This seems to have led to a problem where, when you try to use a function via qAequilbrae dedicated menus (ex. clicking "Open project") or launching something via qgis processing provider :
- Current QGIS instance/session is freezing 
- a new QGIS instance is launching in a new window, 
- when opened, this new instance opens a new instance which opens a new instance, ...
- New instances have error messages about invalid data sources :
```
2023-11-24T16:55:35     CRITICAL    Invalid Data Source : C:\Users\a.evrard.EGIS\AppData\Roaming\QGIS\QGIS3\profiles\1_Classique\python\plugins\qaequilibrae\packages\cpuinfo\cpuinfo.py is not a valid or recognized data source.
2023-11-24T16:55:35     CRITICAL    Invalid Data Source : D:\OSGeo4W\bin\--json is not a valid or recognized data source.
```

Backed-up version of qaequilibrae (before #240) does not have this issue (using import tr from i18n in each file)

@pedrocamargo or @r-akemii, have you seen something when testing the plugin before merging #240 ?

2 examples of working plugins that use transifex : 
- https://github.com/All4Gis/QGISFMV/tree/master/code
- https://github.com/Arpapiemonte/openoise-map/tree/master

r-akemii:
Solves #256. 

* Set tables version to 3.8.0 in requirements.txt
* Fixed the usage of message class in qaequilibrae.py (one cannot use property as input).
* Formatted qaequilibrae,py file with black

### DIFF
--- a/qaequilibrae/message.py
+++ b/qaequilibrae/message.py
@@ -2,12 +2,16 @@ from qgis.PyQt.QtCore import Qt, QCoreApplication
 
 
 class messages:
+
+    @property
+    def first_box_name(self):
+        return self.tr("AequilibraE and other dependencies are not not installed")
+
     @property
     def first_message(self):
-        a = self.tr("AequilibraE and other dependencies are not not installed")
-        b = self.tr("Do you want us to install these missing python packages?")
-        c = self.tr("QGIS will be non-responsive for a couple of minutes.")
-        return f"{a} {b}\r\n{c}"
+        a = self.tr("Do you want us to install these missing python packages?")
+        b = self.tr("QGIS will be non-responsive for a couple of minutes.")
+        return f"{a}\r\n{b}"
 
     @property
     def second_message(self):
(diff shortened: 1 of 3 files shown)


qaequilibrae/message.py
```diff
@@ -2,12 +2,16 @@
 
 
 class messages:
+
+    @property
+    def first_box_name(self):
+        return self.tr("AequilibraE and other dependencies are not not installed")
```
pedrocamargo:
```suggestion
        return self.tr("AequilibraE and other dependencies are not installed")
```

pedrocamargo (approved):

PR #271, opened 2023-12-18
Release of version 0.9.9

pedrocamargo:


qaequilibrae/modules/common_tools/about_dialog.py
```diff
@@ -46,7 +49,7 @@ def __init__(self, iface):
         self.all_items.append([self.tr("AequilibraE Version number"), release_version])
         self.all_items.append([self.tr("GUI version"), b])
         self.all_items.append([self.tr("GUI Repository"), repository])
-        self.all_items.append([self.tr("Minimum QGIS"), "3.14"])
+        self.all_items.append([self.tr("Minimum QGIS"), min_qgis])
```
pedrocamargo, line 54: Good addition!

qaequilibrae/modules/common_tools/about_dialog.py
```diff
@@ -46,7 +49,7 @@ def __init__(self, iface):
...
         self.all_items.append([self.tr("Developers"), par["developers"]])
```
pedrocamargo: We can replace the hardcoded list of developers with something like this:


        import requests
        
        def get_contributors(repo_url):
            repo_name = repo_url.split('/')[-1]
            owner_name = repo_url.split('/')[-2]
            api_url = f"https://api.github.com/repos/{owner_name}/{repo_name}/contributors"
            response = requests.get(api_url)
            if response.status_code == 200:
                contributors = [user['login'] for user in response.json()]
                return [x for x in contributors if "[bot]" not in x]
            else:
                return None
        
        contributors = get_contributors("https://github.com/AequilibraE/qaequilibrae")

qaequilibrae/modules/project_procedures/project_from_osm_dialog.py
```diff
@@ -141,8 +140,8 @@ def signal_handler(self, val):
         elif val[0] == "text":
             self.progress_label.setText(val[1])
         elif val[0] == "finished_threaded_procedure":
-            lines = self.qgis_project.project.network.count_links()
-            nodes = self.qgis_project.project.network.count_nodes()
-            self.report.append(reporter(f"{lines:,} links generated"))
-            self.report.append(reporter(f"{nodes:,} nodes generated"))
```
r-akemii, line 147: @pedrocamargo can you help me with this one? When we select a place from OSM, the network is downloaded and created in AequilibraE. However, when we try to get the number of links and nodes we just downloaded, it raises a warning because we're attempting to access the database in two different ways and it is locked. Do you have any suggestions on how I can solve this one?

r-akemii (approved):

PR #288, opened 2024-01-16
Adds more tests to increase coverage

r-akemii:
In this PR we:

* Removed the duplicated folder _tests_ and moved all its tests to the main test folder (_test_)
* Added _.coveragerc_ file to configure pytest running
* File format with Black!
* Removed _is_testing_ argument from tests and files
* Added new tests:
   * test_create_examples
   * test_create_pro_file
   * test_data_layer_from_dataframe
   * test_load_dataset

**Reaching 74% code coverage!**


qaequilibrae/modules/matrix_procedures/load_dataset_dialog.py
```diff
@@ -170,10 +171,11 @@ def finished_threaded_procedure(self, param):
         self.exit_procedure()
 
     def load_from_aequilibrae_format(self):
-        out_name, _ = GetOutputFileName(self, "AequilibraE dataset", ["Aequilibrae dataset(*.aed)"], ".aed", self.path)
+        if not self._testing:
```
pedrocamargo: Let's NOT add a testing variable

pedrocamargo: def load_from_aequilibrae_format(self):
	out_name, _ = GetOutputFileName(self, "AequilibraE dataset", ["Aequilibrae dataset(*.aed)"], ".aed", self.path)
    self.load_with_file_name(out_name)

#This is the function you test, and ignore the one above
def load_with_file_name(self, out_name):

	try:
		self.dataset = AequilibraeData()
		self.dataset.load(out_name)
	except Exception as e:
		self.error = self.tr(
			"Could not load file. It might be corrupted or not a valid AequilibraE file. {}".format(e.args)

pedrocamargo:
@r-akemii , there are conflicts on this branch

r-akemii:
@pedrocamargo we're with 67% code coverage, almost the 70% we were targeting. I'll add tests in sections with lower coverage in another PR.

test/test_download_extra_packages.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+import os
+from shutil import rmtree
+from pathlib import Path
+
+from qaequilibrae.download_extra_packages_class import DownloadAll
+
+
+@pytest.mark.skip("This test takes a long time to run")
```
pedrocamargo: Should this test be here at all?

test/test_traffic_assignment.py
```diff
@@ -36,13 +36,13 @@ def test_single_class_traffic_assignment(ae_with_project, qtbot):
     dialog.ln_class_name.setText("car")
     dialog.pce_setter.setValue(1.0)
     dialog.chb_check_centroids.setChecked(False)
-    qtbot.mouseClick(dialog.but_add_class, Qt.LeftButton)
+    dialog._create_traffic_class()
```
pedrocamargo: Clicking is not working?

test/test_download_extra_packages.py
r-akemii: Probably not... I'll remove it!

test/test_traffic_assignment.py
r-akemii: Clicking it's working. I changed the clicking actions because I modified the `create_traffic_class` and `add_skimming` actions in the _traffic_assignment_dialog_ file. I'll undo these modifications!

test/conftest.py
```diff
@@ -161,3 +160,153 @@ def create_links_with_matrix():
         layer.dataProvider().addFeatures(features)
 
         QgsProject.instance().addMapLayer(layer)
+
+
+@pytest.fixture
```
pedrocamargo, line 172: Is this fixture being run over and over again?  Do we need to specify that it should only be run once?

pedrocamargo (commented):
Only one question there

test/test_load_plugin.py
pedrocamargo, line 1: Should this file exist?

test/conftest.py
r-akemii: I checked and the fixture is used only in the tests that requested it (distribution procedures and loading result matrices).

pedrocamargo (approved):

PR #297, opened 2024-02-24
BUG: Fix use of backgroundcolor in stacked bandwidths

m-richards:
In the stacked bandwidths dialogue:
image
the up and down buttons don't seem to work on newer versions of QGIS:
image

I'm seeing this on QGIS 3.34 (and I think 3.28 as well). I assume this is to do with the version of Qt (QGIS 3.34 comes with Qt 5.15.3) but I can't seem to find a corresponding changelog where backgroundColor was removed (apparently in the cpp it's just obsolete but not removed: https://doc.qt.io/qt-5/qtablewidgetitem-obsolete.html ?). I'm not sure of backwards compatibility requirements / what promises the qaequilbrae plugin makes, but `background` seems to exist back to Qt 4.2 (https://doc.qt.io/qt-5/qtablewidgetitem.html#background).

(I've also set the default layer in the stacked bandwidths dialogue to be the selected layer, but can revert that if it's not desired)

### DIFF
--- a/qaequilibrae/modules/gis/create_bandwidths_dialog.py
+++ b/qaequilibrae/modules/gis/create_bandwidths_dialog.py
@@ -40,6 +40,7 @@ class CreateBandwidthsDialog(QDialog, FORM_CLASS):
         # layers and fields        # For adding skims
         self.mMapLayerComboBox.layerChanged.connect(self.add_fields_to_cboxes)
         self.mMapLayerComboBox.setFilters(QgsMapLayerProxyModel.LineLayer)
+        self.mMapLayerComboBox.setCurrentText(self.iface.activeLayer().name())
 
         self.ab_FieldComboBox.currentIndexChanged.connect(partial(self.choose_a_field, "AB"))
         self.ba_FieldComboBox.currentIndexChanged.connect(partial(self.choose_a_field, "BA"))
--- a/qaequilibrae/modules/gis/create_bandwidths_dialog.py
+++ b/qaequilibrae/modules/gis/create_bandwidths_dialog.py
@@ -40,7 +40,7 @@ class CreateBandwidthsDialog(QDialog, FORM_CLASS):
         # layers and fields        # For adding skims
         self.mMapLayerComboBox.layerChanged.connect(self.add_fields_to_cboxes)
         self.mMapLayerComboBox.setFilters(QgsMapLayerProxyModel.LineLayer)
-        self.mMapLayerComboBox.setCurrentText(self.iface.activeLayer().name())
+        self.mMapLayerComboBox.setLayer(self.iface.activeLayer())
 
         self.ab_FieldComboBox.currentIndexChanged.connect(partial(self.choose_a_field, "AB"))
         self.ba_FieldComboBox.currentIndexChanged.connect(partial(self.choose_a_field, "BA"))
--- a/qaequilibrae/modules/gis/create_bandwidths_dialog.py
+++ b/qaequilibrae/modules/gis/create_bandwidths_dialog.py
@@ -183,11 +183,11 @@ class CreateBandwidthsDialog(QDialog, FORM_CLASS):
 
         # if we clicked the "remove button"
         if column == 0 and row < self.tot_bands - 1:
-            low_cl = self.bands_list.item(row + 1, 2).backgroundColor()
+            low_cl = self.bands_list.item(row + 1, 2).background()
             low_ab = self.bands_list.item(row + 1, 0).text()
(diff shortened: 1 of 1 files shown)


pedrocamargo:
Excellent contribution, @m-richards . Thank you!

pedrocamargo (approved):

m-richards:
Separately I was wondering if there's any appetite to add some functionality to make it easier to inspect which colours correspond to which fields after creating stacked bandwidths. I've been doing this a few times recently and have found myself taking screenshots of the creation dialogue as reference when using the layer. Some thoughts I had on this were to either
* (optionally) create a dummy "legend" layer e.g. this in reference to my first screenshot:
image
* Maintaining some state in order to pre-populate the selected flow fields and colours when the stacked bandwidth dialogue is opened for an already styled field, so you can use the dialogue itself as a reminder. With this there's at least a couple of approaches: 
    * when the stacked bandwidth tool is run for a layer you store the state of the bandwidth in memory and restore it if the field is selected again in the stacked bandwidth tool (but this gets out of sync if someone manually changes the colours / styling after the fact (not a concern I care about, but it's probably still valid), the project is saved and re opened, etc.
    * When the stacked bandwith tool is run, try and detect if the layer has already been styled by qaequilibrae (i.e. has the AB BA line structure per band, dynamic width expression matches what is set by qaequilibrae). This is likely pretty convoluted but avoids edge cases of colours getting out of sync if they're manually edited after the fact.

If there's interest I'd be happy to contribute a PR along one of these lines (or other suggestions), but appreciate this isn't exactly core functionality for the plugin, so perhaps it's overkill.

pedrocamargo:
That would be excellent to have indeed.  

I have recently added this (idea  No. 1) to Polaris, so I do have some code I could port to QAequilibraE if you'd like.

Maintaining states in QAequilibraE is something I have long wanted to do, and my plan was to add a yaml/json to the temp folder in order to do that.  I currently do something in those lines in QPolaris, but that's just for the opening of projects...

pedrocamargo:
The QPolaris code borrowed a lot from QAequilibraE, so the code below should be helpful

color legend (in that case it is not for stacked, but rather for a color ramp)


        legend_steps = 5
        color_layer = QgsVectorLayer("LineString?crs=4326", f"{self.color_results.metric_name} (Color)", "memory")
        symbol = QgsSymbol.defaultSymbol(color_layer.geometryType())
        symbol.setWidth(2)
        renderer = QgsRuleBasedRenderer(symbol)

        ref_style = QgsStyle().defaultStyle()
        color_ramp = ref_style.colorRamp(self.color_ramp)
        max_legend = self.maxcolor / self.dsp_color_thresh.value()
        for interval in range(legend_steps + 1):
            val = interval / legend_steps
            rule = renderer.rootRule().children()[0].clone()
            label = f"{(val * max_legend):,.2f}" if max_legend < 100 else f"{int(val * max_legend):,}"
            rule.setLabel(label)
            rule.symbol().setColor(color_ramp.color(val))
            renderer.rootRule().appendChild(rule)

Thickness legend



        legend_steps = 5
        width_layer = QgsVectorLayer("LineString?crs=4326", f"{self.results.metric_name} (Width)", "memory")
        symbol = QgsSymbol.defaultSymbol(width_layer.geometryType())
        symbol.setWidth(0.001)
        symbol.setColor(QColor("#000000"))

        renderer = QgsRuleBasedRenderer(symbol)
        max_legend = self.maxwidth / self.dsp_thick.value()
        for interval in range(legend_steps + 1):
            val = interval / legend_steps
            rule = renderer.rootRule().children()[0].clone()
            label = f"{(val * max_legend):,.2f}" if max_legend < 100 else f"{int(val * max_legend):,}"
            rule.setLabel(label)
            rule.symbol().setWidth(val * self.__map_width)
            renderer.rootRule().appendChild(rule)

        # remove first child
        renderer.rootRule().removeChildAt(0)
        width_layer.setRenderer(renderer)
        QgsProject.instance().addMapLayer(width_layer)
        width_layer.triggerRepaint()

m-richards:
> That would be excellent to have indeed.
> 
> I have recently added this (idea No. 1) to Polaris, so I do have some code I could port to QAequilibraE if you'd like.
> 
> Maintaining states in QAequilibraE is something I have long wanted to do, and my plan was to add a yaml/json to the temp folder in order to do that. I currently do something in those lines in QPolaris, but that's just for the opening of projects...

Thanks for the starting point snippet - I'll go with 1 for now then - it's the simplest starting point and it's not incompatible with persisting state later either.

PR #301, opened 2024-03-10
ENH: Add legend layer for stacked bandwidths

m-richards:
Finally came back to this, WIP for now. Have something working for the color legend side of things, nothing for thickness yet. This is a bit of a silly example mixing solid colours and colour ramps;
image
but it shows that I've got categorical legend entries working, as well as graduated entries for the color ramp option, which also supports different color ramps and data for the AB and BA directions (unticking dual fields). I've added things hierarchically into the legend rules as I think that makes the most sense.

Edit: Also now with a thickness legend:
image

There are a couple of TODOs of things in the code I'm not sure about / could be cleaned up.
Also was wondering about how to expose this. Currently I'm always generating the legends, but I could see an argument there should be a checkbox to turn legend generation off?


qaequilibrae/modules/gis/create_bandwidths_dialog.py
```diff
@@ -295,16 +319,16 @@ def add_bands_to_map(self):
         QgsExpressionContextUtils.setProjectVariable(QgsProject.instance(), "aeq_band_max_value", max_value)
         QgsExpressionContextUtils.setProjectVariable(QgsProject.instance(), "aeq_band_spacer", float(space_size))
         QgsExpressionContextUtils.setProjectVariable(QgsProject.instance(), "aeq_band_width", band_size)
-        max_value = "@aeq_band_max_value"
-        space_size = "@aeq_band_spacer"
-        band_size = "@aeq_band_width"
+        max_value_aeq_var = "@aeq_band_max_value"
```
m-richards, line 332: renamed so that I can use the numerical value of these variables again below.

qaequilibrae/modules/gis/create_bandwidths_dialog.py
```diff
@@ -369,11 +393,129 @@ def add_bands_to_map(self):
...
+        self.build_color_legend(self.layer.name(), bands=bands)
 
         self.layer.renderer().symbol().deleteSymbolLayer(0)
         self.layer.triggerRepaint()
         self.exit_procedure()
 
     def exit_procedure(self):
         self.close()
+
+    @classmethod
```
m-richards, line 426: these could just be functions in the module, as classmethods for now.

r-akemii (approved):

PR #305, opened 2024-03-22
Adds PT visualization tool

r-akemii:
In this PR we add:

* a PT database viewer with layer styles;
* the documentation;
* the tests.


pedrocamargo:
Shouldn't pattern_id be in the same table as the routes?

image

pedrocamargo (approved):

PR #318, opened 2024-04-09
Fixes OMX usage in Traffic Assignment

r-akemii:
This PR adds a verification of existence of _openmatrix_, and in case it does not exist, adds the description "_(OMX not available)_".

Currently, only \*.aem files are being used in Traffic Assignment because all \*.omx files are considered as not available.


qaequilibrae/modules/paths_procedures/traffic_assignment_dialog.py
```diff
@@ -20,6 +21,9 @@
 from qaequilibrae.modules.common_tools import ReportDialog
 from qaequilibrae.modules.common_tools import standard_path
 
+spec = iutil.find_spec("openmatrix")
```
pedrocamargo: openmatrix is now a hard requirement. Let's remove the checks and extra stuff and assume that, if the user got all the way here, they will be able to use OMX

r-akemii: Same here. I removed the "has_omx" pieces of code from traffic_assignment_dialog (it should be fine now)

pedrocamargo (approved):

PR #320, opened 2024-04-11
Updates project layer creation

r-akemii:
This PR:

* creates a function (_update_project_layers_) that can be reused in the code
* modifies the code in the gtfs_importer _execute_import_ function to delete and update the layers in the project if a GTFS is imported. 

This modification was done because when importing a GTFS to a project, and immediately trying to explore it using _Explore Transit_, an error regarding the existence of stops, routes, and pattern layers was displayed. This way we remove the error, and make the user experience a little better by updating the _Geo layers_ tab, which would only be updated the next time we opened the project.

### DIFF
--- a/qaequilibrae/modules/menu_actions/load_project_action.py
+++ b/qaequilibrae/modules/menu_actions/load_project_action.py
@@ -39,6 +39,9 @@ def _run_load_project_from_path(qgis_project, proj_path):
             else:
                 raise e
 
+    update_project_layers(qgis_project)
+
+def update_project_layers(qgis_project):
     curr = qgis_project.project.conn.cursor()
     curr.execute("select f_table_name from geometry_columns;")
     layers = [x[0] for x in curr.fetchall()]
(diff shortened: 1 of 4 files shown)


qaequilibrae/modules/menu_actions/load_project_action.py
```diff
@@ -39,6 +39,9 @@ def _run_load_project_from_path(qgis_project, proj_path):
             else:
                 raise e
 
+    update_project_layers(qgis_project)
+
+def update_project_layers(qgis_project):
     curr = qgis_project.project.conn.cursor()
```
pedrocamargo, line 45: How does this work?  isn't this connection to the project_database?

r-akemii: This function _update_project_layers_ is the one responsible for populating the geo layers tab. I initially thought of re-using it because in both the project import and in gtfs_import there is an open AequilibraE project, but I can come up with something more adequate rather than this. There is a bug just after importing a GTFS feed that I cannot use the explore transit because the GTFS layers are not in the layers dict yet, and they do not appear in the geo layers tab.

pedrocamargo (approved):

PR #321, opened 2024-04-11
Updates on Data Visualization menu

r-akemii:
This PR:

* Reorders tab displaying for _ui_project_data_
* Fixes the controls for _display_aequilibrae_formats_dialog.py_ for a situation in which the user closes the GetOutputFileName window. Before it raised an error and the _ui_vector_loader_ was not responsive. Now it still raises an error, but the UI is responsive, so the user can click once again to load the data or close the window. There is also a modification to allow the users to view the content of OMX, AEM or AED files using QAequilibraE without opening a project.
* Reflects these modifications in test files and other dialogs
* Adds test for LoadDatasetDialog class

### DIFF
--- a/qaequilibrae/modules/matrix_procedures/display_aequilibrae_formats_dialog.py
+++ b/qaequilibrae/modules/matrix_procedures/display_aequilibrae_formats_dialog.py
@@ -178,10 +178,15 @@ class DisplayAequilibraEFormatsDialog(QtWidgets.QDialog, FORM_CLASS):
             self.add_matrix_parameters(idx, core)
             self.format_showing()
 
-    def export(self):
-        new_name, file_type = GetOutputFileName(
+    def csv_file_path(self):
+        new_name, _ = GetOutputFileName(
             self, self.data_type, ["Comma-separated file(*.csv)"], ".csv", self.data_path
         )
+        return new_name
+
+    def export(self):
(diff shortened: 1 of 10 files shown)


qaequilibrae/modules/matrix_procedures/display_aequilibrae_formats_dialog.py
```diff
@@ -214,3 +210,21 @@ def add_matrix_parameters(self, idx, field):
         self.data_to_show.matrix_view = np.array(self.omx[field])
         self.data_to_show.index = np.array(list(self.omx.mapping(idx).keys()))
         self.data_to_show.matrix[field] = self.data_to_show.matrix_view[:, :]
+
+    def get_file_name(self):
+        formats = ["Aequilibrae matrix(*.aem)", "Aequilibrae dataset(*.aed)"]
+
+        dflt = ".aem"
+        if has_omx:
```
pedrocamargo: Let's assume it DOES have OMX

r-akemii: I removed the "has_omx" from display_aequilibrae_formats_dialog and load_project_data (files modified in this PR)

pedrocamargo (approved):

PR #330, opened 2024-04-29
Save aequilibrae project directly from qgis

### LINKED ISSUE #323: [ENH] Save AequilibraE project directly from QGIS
Saving an AequilibraE project to a QGIS project is triggered by a built-in QAequilibraE menu command. Ideally, we could save AequilibraE project to a QGZ file directly using the saving button rather than clicking on the menu.

r-akemii:
This PR modifies the _qaequilibrae.py_ file to allow saving AequilibraE project directly from QGIS, without having to activate the command via plugin.

Related tests and documentation files were also updated and/or removed.


pedrocamargo:
@r-akemii , the only thing we are missing is a clear note in the documentation saying that saving a QGIS project automatically saves the AequilibraE project. That is too good of a feature to not brag about

pedrocamargo (approved):

PR #354, opened 2024-06-17
[WIP] matrix calculator, matrix import/updt, faster project creation

Art-Ev:
Renaming processings to handle menus order & add datetime to assignment result name for easier re-running without editing yaml config file


qaequilibrae/modules/processing_provider/Add_connectors.py
```diff
@@ -78,10 +78,10 @@ def displayName(self):
         return self.tr("Add centroid connectors")
 
     def group(self):
-        return self.tr("Model Building")
+        return self.tr("01-Model Building")
```
r-akemii, line 143: I don't like numbering the menus. I get that it's easier to understand the step-by-step as in the AequilibraE menu, but no other providers (QGIS and GDAL, for instance) numbered their menus and it could create another translatable string. When it comes to translatable strings, I think we have to keep it simple and reuse as much as we can.

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
```diff
@@ -0,0 +1,165 @@
...
+                multiLine=False, 
+                defaultValue="Value"),
+        ]
+
+    def processAlgorithm(self, parameters, context, model_feedback):
+        feedback = QgsProcessingMultiStepFeedback(3, model_feedback)
+
+        # Checks if we have access to aequilibrae library
+        if iutil.find_spec("aequilibrae") is None:
+            sys.exit(self.tr("AequilibraE module not found"))
```
r-akemii, line 74: Is this necessary? Currently, the provider is only installed with the QAequilibraE plugin so aequilibrae is a must-have by default.

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
```diff
@@ -0,0 +1,165 @@
...
+        mat.close()
+        del agg_matrix, matrix, m
+
+        feedback.pushInfo(" ")
+        feedback.setCurrentStep(3)
+
+        return {"Output": f"{mat.name}, {mat.description} ({file_name})"}
+
+    def name(self):
+        return self.tr("Add a matrix from a layer to an existing aem file")
```
r-akemii: Can we shorten this name? Something like _AEM from layer_, maybe?

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
```diff
@@ -0,0 +1,165 @@
...
+        if order == 1:
+            return self.tr("Save a layer to an existing *.aem file. Notice that:")
+        elif order == 2:
+            return self.tr("- the original matrix stored in the layer needs to be in list format")
+        elif order == 3:
+            return self.tr("- origin and destination fields need to be integers")
+        elif order == 4:
+            return self.tr("- value field can be either integer or real")
+        elif order == 5:
+            return self.tr("- if matrix_core is already existing, then it will be updated and previous data will be lost")
```
r-akemii: grammar? _if matrix_core already exists, it will be updated and previous data will be lost_

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
```diff
@@ -0,0 +1,165 @@
...
+                type=QgsProcessingParameterField.Numeric,
+                parentLayerParameterName="matrix_layer",
+                allowMultiple=False,
+                defaultValue="value",
+            )
+        )
+        self.addParameter(
+            QgsProcessingParameterFile(
+                "matrix_file",
+                self.tr("Matrix file"),
```
r-akemii: As this is related to an existing AEM file, maybe we could call it _Existing AEM file_?

qaequilibrae/modules/processing_provider/create_matrix_from_layer.py
```diff
@@ -144,30 +144,30 @@ def processAlgorithm(self, parameters, context, model_feedback):
...
+        del agg_matrix, matrix, m
 
         feedback.pushInfo(" ")
         feedback.setCurrentStep(3)
 
         return {"Output": f"{mat.name}, {mat.description} ({file_name})"}
 
     def name(self):
-        return self.tr("Import matrices")
+        return self.tr("Create an aequilibrae matrix file (.aem) from a layer")
```
r-akemii: Can we shorten this name?

qaequilibrae/modules/processing_provider/assign_from_yaml.py
```diff
@@ -112,7 +114,7 @@ def processAlgorithm(self, parameters, context, model_feedback):
         # Saving outputs
         feedback.pushInfo(self.tr("Saving outputs"))
         feedback.pushInfo(str(assig.report()))
-        assig.save_results(params["result_name"])
+        assig.save_results(params["result_name"]+dt.now().strftime('_%Y-%m-%d_%Hh%M'))
```
r-akemii: Is this necessary? If I'm not mistaken, when we run an assignment in aequilibrae, its results are saved into results.database with a timestamp. Also in the aequilibrae.log for the project, the assignment info is stored too.

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
```diff
@@ -0,0 +1,165 @@
...
+        feedback.pushInfo(self.tr("Importing layer"))
+        layer = self.parameterAsVectorLayer(parameters, "matrix_layer", context)
+        cols = [origin, destination, value]
+        datagen = ([f[col] for col in cols] for f in layer.getFeatures())
+        matrix = pd.DataFrame.from_records(data=datagen, columns=cols)
+        feedback.pushInfo("")
+        feedback.setCurrentStep(1)
+
+        # Getting all zones
+        all_zones = np.array(sorted(list(set(list(matrix[origin].unique()) + list(matrix[destination].unique())))))
```
r-akemii: This is a bit confusing. You could re-write this as something like:

`a = np.array([1, 2, 5, 6, 7], dtype=np.int32)`
`b = np.array([1, 3, 5, 8, 9], dtype=np.int32)`

`np.unique(np.concatenate((a, b)))`
`# returns: array([1, 2, 3, 5, 6, 7, 8, 9])`

It's probably not the best way to do this, but it's easier to read, returns an numpy array, and sorts the data.

qaequilibrae/modules/processing_provider/Add_connectors.py
Art-Ev: QGIS and GDAL don't have any methodological nor computing order between processes unlike AequilibraE. I also think that if we keep name from lefts menus we should also keep the same display order and here only the alphabetised order is available...
Next commit will separate the numbering from menu real "name" to make fewer different strings when translating ;)

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
Art-Ev: Until aequilibrae library is directly send with the plugin files yes it is : when starting, qgis is checking processing provider scripts. If one of them gives an error (for ex if library import fails) : whole plugin is desactivated to prevent errors. So, at first installation this prevents the plugin from downloading the aequilibrae library (désactivated by QGIS before) 😢.

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
Art-Ev: Other proposal in next commit

qaequilibrae/modules/processing_provider/assign_from_yaml.py
Art-Ev: It's for skim matrix files (stored in one unique folder and don't have datetime in file name so only one file at a time is possible). In next commit I'll updt to set this has facultativ

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
Art-Ev: I think it was taken from a very old aequilibrae example... (change in next commit after perf checks)

qaequilibrae/modules/processing_provider/add_matrix_from_layer.py
r-akemii: Wow, I didn't know that.

r-akemii:
@Art-Ev by the way, do you have any sample files you're using to test the provider to send me?

Art-Ev:
> @Art-Ev by the way, do you have any sample files you're using to test the provider to send me?

Not yet, it's one shots every time... But we have somebody with whom we'll be working on this over the summer: a full model creation and running using only qgis processes from the provider with precalculated results to validate future updates and future new functions. Goal is to have one QGIS modeler that we can juste run with one button and check that most functions work as expected.

r-akemii:
@Art-Ev I did a super mess in your branch / commit :( Je suis vraiment désolée de vous dérranger

Art-Ev:
> @Art-Ev I did a super mess in your branch / commit :( Je suis vraiment désolée de vous dérranger

Seen that trying to push last commit 😋 I think it's good now (had lot of errors before but final files in "files changes" tab seem to be correct at the end)

qaequilibrae/modules/processing_provider/Add_connectors.py
```diff
@@ -1,13 +1,17 @@
...
 import sys
+from os.path import join
+import pandas as pd
 
-from qgis.core import QgsProcessingMultiStepFeedback, QgsProcessingAlgorithm
+from qgis.core import QgsProcessingMultiStepFeedback, QgsProcessing, QgsProcessingAlgorithm
 from qgis.core import QgsProcessingParameterFile, QgsProcessingParameterNumber, QgsProcessingParameterString
+from qgis.core import QgsFeature, QgsVectorLayer, QgsDataSourceUri
+
+import processing
```
r-akemii, line 10: Hey @Art-Ev I want to check with you about this import because the test I wrote is failing with these changes. I tried to import this module but I received an error. Would you mind checking this for me, please?

Art-Ev: Hey, will check that, which version of QGIS are you using ? (no error on 2 different versions on my computer nor on another person's...)

Art-Ev:
@r-akemii, we've made some tests and only have a last problem : PT assign and map matching are buggy on our side. (errors during graph creation). Can you check that ? If you see no problem on your side we'll organize a call to send you our test files.

r-akemii:
@Art-Ev I can definitely investigate the bug! Would you mind describing it and sending me the GTFS files you were using?

I have a hunch that it can be related to a bug Pedro fixed in AequilibraE in release 1.1.0.

Art-Ev:
> @Art-Ev I can definitely investigate the bug! Would you mind describing it and sending me the GTFS files you were using?
> 
> I have a hunch that it can be related to a bug Pedro fixed in AequilibraE in release 1.1.0.

Will make that tomorrow ;)

r-akemii:
@Art-Ev I'm not sure if I can reach you on Teams, but I'll contact you tomorrow during your working hours!

qaequilibrae/modules/processing_provider/Add_connectors.py
r-akemii: Hey @Art-Ev I'm updating the tests based on the changes in your branch and I still have problems importing `processing`. My container is using QGIS LTR (3.34.13 IMO) . Can you please check this import for me?

PR #355, opened 2024-06-21
Fixes plugin i18n strings

### LINKED ISSUE #350: Verify translatable strings
It would be ideal to:

* check if all strings marked as translatable are indeed necessary;
* check if all localized strings are being translated in Transifex;
* check if all strings translated with Transifex are "placed" in the right spot;
* check if there are any missing strings to be localized;
* update Transifex translation documentation;
* add a subsection with formatting good practices for translatable strings.

r-akemii:
This PR aims to check the translation status on qaequilibrae, and update the documentation.

In this PR we:

- checked the position of all translatable strings in the plugin
- removed unnecessary strings marked as translatable
- updated the plugin internationalization documentation


r-akemii:
@Art-Ev Currently we have translations in FR and FR-FR. Do you believe they're both necessary or is just FR translation ok? I can set up the configurations to consider just the first one. I'm asking you because Brazilian Portuguese is very different from European Portuguese, thus it makes sense to have both translations IMO.

pedrocamargo (approved):

PR #358, opened 2024-07-29
Fixes installation on python 312

r-akemii:
In this PR we fix the installation issues related to Python 3.12. Thus, we:

* Modified the requirements.txt and requirements-test.txt files;
* Modified the extra classes download file to remove existing packages in python 312;
* Added a new installation message warning users that qaequilibrae is running only on python 312 for Windows users;
* Modified the devworkflow file to use qgis 3.34.7 rather than release-3_34 (version 3.34.6 used python 310);
* Updates the release version to 1.0.3

I manually installed the artifact ZIP-file in versions 3.34.6, 3.36.0, 3.38.0, 3.38.1, and the warning messages and installation error messages seems to be popping out when they should.


qaequilibrae/download_extra_packages_class.py
```diff
@@ -13,32 +13,36 @@ class DownloadAll:
...
+        on_latest = version >= (3,12)
         with open(self.file, "r") as fl:
             lines = fl.readlines()
 
         reps = []
         for line in lines:
             reps.extend(self.install_package(line.strip(), on_latest))
 
+        if version.micro > 3:
+            self.must_remove.append("pyogrio")
```
pedrocamargo: pyogrio does not come with newer versions of python, but rather with newer versions of GeoPandas.  Why can't we just add to the list of things that would have to be removed anyways?

qaequilibrae/message.py
```diff
@@ -25,6 +25,12 @@ def third_message(self):
     @property
     def fourth_message(self):
         return self.tr("Without installing the packages, the plugin will be mostly non-functional")
+    
+    @property
+    def messsage_five(self):
+        a = self.tr("QAequilibraE requires Python 3.12.")
+        b = self.tr("Please install an appropriate version to make it work")
```
pedrocamargo: "Install an appropriate version" does NOT help.  Let's be specific.  "Please install version 3.34.6+ or 3.36.6+" (or whatever the versions are).

This also should only apply to Windows, right?

qaequilibrae/requirements.txt
```diff
@@ -1,9 +1,5 @@
...
 aequilibrae==1.0.1
 ortools
-cython
-numexpr
-py-cpuinfo
-tables
-openmatrix
-fastparquet
\ No newline at end of file
+absl-py
```
pedrocamargo, line 3: Why?

test/requirements-test.txt
```diff
@@ -1,9 +1,8 @@
-numpy
-geopandas
 pandas
-pyaml
-pyproj
-requests
+geopandas
 rtree
-scipy
 shapely
```
pedrocamargo: shapely and Pandas can be removed, as they are both dependencies of GeoPandas

test/test_qaequilibrae_menu_with_project.py
```diff
@@ -298,6 +299,7 @@ def test_gtfs_explorer_no_gtfs(ae_with_project):
     ), "Level 3 error message is missing"
 
 
+@pytest.mark.skip
```
pedrocamargo: Why?

test/requirements-test.txt
r-akemii: I'll remove geopandas too, because it's one of aequilibrae dependencies, and now it's being installed with all dependencies.

.github/workflows/DevWorkflow.yml
```diff
@@ -10,27 +10,18 @@ jobs:
...
 
-      - name: Install dependencies for Latest
-        if: ${{ (matrix.container == 'qgis/qgis:latest')}}
+      - name: Install dependencies
         run: |
           python3 -m pip install setuptools --break-system-packages
           python3 -m pip install -r test/requirements-test.txt --break-system-packages
-          python3 -m pip install -r ci/requirements-dev.txt --break-system-packages
-          python3 ./ci/dependency_installation.py
+          python3 -m pip install aequilibrae==1.0.1 absl-py immutabledict protobuf --break-system-packages
```
pedrocamargo, line 23: Voltou com a história de instalar pro CI de um jeito e no teste de outro. Não pode, né?

PR #361, opened 2024-08-12
Adds test on Windows

r-akemii:
In this PR we add a new _.github/workflow_ to test the QAequilibraE plugin on the QGIS Windows installation for versions 3.34.6, LTR, and Latest.

We also modified _download_extra_packages_class.py_ to be executed with the command line and modified the _.gitignore_ file to ignore the packages installed.

**Note**: Currently we're not running pytest within this workflow.


r-akemii:
@pedrocamargo I didn't manage to configure pytest to run with the qaequilibrae installation in qgis. The best I can do now is test if the missing packages are being installed in the right place, and if the installation is not raising any errors. I'll try to find the right way to run the tests, but I decided to give it a rest because I was going nowhere.

pedrocamargo (approved):

PR #374, opened 2024-10-04
Fixes installation on Python 3.12.6

r-akemii:


r-akemii:
@pedrocamargo I did some tests on my machine, removing / not removing the duplicated installation packages. I didn't notice any breakdowns in installation, but I don't think they're necessary. What I found out though is that QAequilibraE installation is not working on LTR >= 3.34.10 and latest >= 3.38.2, and it's possibly related to a conflict with one of pyarrow dependencies on Python 3.12.6. Installing pyarrow without dependencies and with a different version in Python 3.12.6 and 3.12.4 works fine.

PR #380, opened 2024-10-24
Integrate with new AequilibraE docs

r-akemii:
In this PR we have modified the structure of the QAequilibraE documentation.

* Updated documentation workflow to a new Python version and added a PDF generator for the docs
* Updated .gitignore
* Updated _conf.py_
* Renamed the index file to _qgis.rst_ based on the new AWS folder structure
* Added custom configuration to page headers and sidebars
* Updated contents in menus in detail and software development
* Removed contributors.rst file
* Added a folder with a \*.rst file with configurations to the Latex PDF builder


r-akemii:
@pedrocamargo I didn't manage to add a second data level in the primary sidebar as you suggested because the headers are all on the same page. I don't think it's worth having several "short pages" just to add this level. What I did though was display the level 2 headers in the "initial" menu file. What do you think? 

image

pedrocamargo (approved):

PR #401, opened 2025-02-04
Add Select Link Analysis

r-akemii:
In this PR we add Select Link Analysis tool to Paths and Assignments.

We:

- Updated traffic_assignment_ui and traffic_assignment_dialog to run select link analysis
- Added a conditional test for select link analysis (based on the bug identified in AequilibraE)
- Updated documentation with an example for the new feature


qaequilibrae/modules/paths_procedures/forms/ui_traffic_assignment.ui
r-akemii, line 1: Current tab view
image

pedrocamargo: What's the difference between "Add query" and "build query"? It is not clear at all.

I would do something like "Add to query" and "Save query." It is a lot clearer.

pedrocamargo (approved):

PR #418, opened 2025-03-19
Add Docker file to speed up Windows Release test

r-akemii:
**See comments below**


r-akemii:
### FYI

Pedro and I discussed about this pull on march 2025. We were able to create an useful container with a QGIS LTR installation on Windows but what we noticed is that we cannot use this container (even if its working and in GHCR) due to an incompatibility across Windows containers and GitHub:

1) we cannot use `windows-latest` to run the container because containers cannot be run on windows;
2) and we cannot use `linux-latest` to run the container because its a windows container (incompatible)

This is an old issue and so far, there's no solution. Check https://github.com/actions/runner/issues/904 for more info.

PR #450, opened 2025-06-06
Add run module

r-akemii:
In this PR we incorporate AequilibraE's run module into the QGIS widget and the processing provider, adding the necessary tests and documentation.


qaequilibrae/modules/common_tools/auxiliary_functions.py
```diff
@@ -113,3 +114,11 @@ def model_area_polygon(poly) -> Union[Polygon, int]:
...
     zones = poly.to_crs(4326)
     return zones.union_all().buffer(tol).buffer(-tol), 4326
+
+
+def list_func(filepath):
+    """Lists the existing functions in a file."""
+    with open(filepath, "r", encoding="utf-8") as f:
+        file_content = f.read()
+    tree = ast.parse(file_content)
+    return [node.name for node in ast.walk(tree) if isinstance(node, ast.FunctionDef)]
```
Jake-Moss: Do we want to be parsing the AST here? This is also looking for the function definitions specifically, but it's possible to use a class, or imports like this

```python
from .foo import bar as bar
```

IMO we should use the `parameters.yml` run header as the source of truth for what functions are available. If a function isn't listed then we don't display it.

qaequilibrae/modules/processing_provider/run_module.py
```diff
@@ -0,0 +1,90 @@
...
+                    )
+                )
+                return
+            else:
+                self.project = aeq_plugin.project
+
+            # Check if there's an open project and fetch its information
+            if self.project:
+                self.filepath = join(self.project.project_base_path, "run/__init__.py")
+                self.items = list_func(self.filepath)
```
Jake-Moss: Can we use `self.items = list(self.project.parameters["run"].keys())` here instead? If it's more useful I can make the run module only contain attributes from the `parameters.yml` instead of replacing attributes in the real module, that way you can just get the module (which would return as a `namedtuple`) then iterate over it with `._fields`

qaequilibrae/modules/project_procedures/run_module_dialog.py
```diff
@@ -0,0 +1,48 @@
...
+
+        self.items = list_func(self.filepath)
+        self.cob_function.addItems(self.items)
+
+        self.but_run.clicked.connect(self.run)
+
+    def run(self):
+        spec = spec_from_file_location("__init__", self.filepath)
+        module = module_from_spec(spec)
+        spec.loader.exec_module(module)
```
Jake-Moss: Same here

qaequilibrae/modules/processing_provider/run_module.py
```diff
@@ -0,0 +1,90 @@
...
+            QgsMessageLog.logMessage(f"Error checking AequilibraE project: {str(e)}")
+
+    def processAlgorithm(self, parameters, context, feedback):
+        feedback.pushInfo("Get logger")
+        logger = get_logger()
+
+        feedback.pushInfo("Define spec from location")
+        spec = spec_from_file_location("__init__", self.filepath)
+        module = module_from_spec(spec)
+        spec.loader.exec_module(module)
```
Jake-Moss: Can we use the `project.run` property here? It handles the module loader so we're not duplicating this snippet. I made an error with it a couple weeks ago and have put a PR up to fix now https://github.com/AequilibraE/aequilibrae/pull/674

Otherwise this snippet also needs to add the module to `sys.modules` for relative imports to work

Jake-Moss (dismissed):
This UI looks solid, unfortunately it doesn't work on my test network because there are no function definitions in the `__init__.py` file, only imports, so the "Available functions" drop down is empty.

If there's changes I can make to the `project.run` method to make this easier and more robust let me know

pedrocamargo (approved):

PR #472, opened 2025-07-10
Fix bug on route choice

r-akemii:
In this PR we fix a bug when using "_ab"/"_ba" fields as utility function and rename cost field to "\_\_utility\_\_" to avoid conflict with a homonym field that might exist.

We modified the graph creation and some of the functions to prevent crashes in QGIS, and modified a test to assert these changes in code.


pedrocamargo (approved):
Excellent work!

PR #475, opened 2025-07-14
Add requirements to run procedures

r-akemii:
In this PR, we modified the workflow and added extra messages and logs to the Run Procedures module to handle cases where external requirements need to be installed.


qaequilibrae/message.py
```diff
@@ -32,5 +32,23 @@ def messsage_five(self):
...
+    @property
+    def rp_message(self):
+        """Message for run procedures"""
+        a = self.tr("There are missing requirements to run the procedures.")
+        b = self.tr("Do you want us to install these missing Python packages?")
+        c = self.tr("Without installing the packages, you cannot use 'Run Procedures'.")
+        return f"{a}\r\n{b}\r\n{c}"
+
+    @property
+    def rp_error(self):
```
pedrocamargo: The project may not have additional requirements, in which case we should ignore it and just log it

r-akemii: @pedrocamargo I ran some tests here with `matplotlib`, which is in the QGIS default installation, with no `requirements.txt`. The procedure is properly executed, without displaying this message. The `rp_error` should only appear if a procedure requires external libraries listed in the `requirements.txt` that don't exist.

I'll move the messages from `messages.py` to `run_module_dialog.py` for ease of debugging, and I'll run more tests here just to be sure there's nothing wrong with the workflow.

qaequilibrae/modules/project_procedures/run_module_dialog.py
```diff
@@ -1,29 +1,90 @@
...
+                        stderr=subprocess.STDOUT,
+                        universal_newlines=True,
+                    )
+                    ret = process.stdout.readlines()
+                    for line in ret:
+                        QgsMessageLog.logMessage(line, level=Qgis.MessageLevel.Info)
+
+                    # Check process output
+                    exit_code = process.wait()
+                    if exit_code != 0:
```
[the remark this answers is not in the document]
r-akemii: I already tested with `run` but it didn't work,

pedrocamargo (approved):

PR #479, opened 2025-07-18
Add scenarios to QAequilibraE

r-akemii:
In this PR we add the AequilibraE scenario tool into QAequilibraE.

We:

* Added a create scenario class and UI
* Modified the scenario comparison UI and workflow to allow comparison between data from different scenarios
* Added a layer from geodataframe function
* Updated database connections with the new AequilibraE standard
* Updated tests and docs
* Modified user communication using messaging logs
* Added deprecation warnings for AEM files
* Modified exporting matrix format to OMX rather than AEM


qaequilibrae/modules/matrix_procedures/load_project_data.py
```diff
@@ -59,28 +55,28 @@ def display_matrix(self):
...
+        self.matrices = list_matrices(self.project)
 
         self.matrices_model = PandasModel(self.matrices)
         self.list_matrices.setModel(self.matrices_model)
 
     def update_matrix_table(self):
         matrices = self.project.matrices
         matrices.update_database()
-        with commit_and_close(database_connection("network")) as conn:
+        with self.project.db_connection_spatial as conn:
```
Jake-Moss: Does this need to be spatial?
```suggestion
        with self.project.db_connection as conn:
```

qaequilibrae/modules/matrix_procedures/results_lister.py
```diff
@@ -1,27 +1,29 @@
...
         databases.append(df)
 
-    if isfile(join(project_base_path, "public_transport.sqlite")):
-        with read_and_close(database_connection("transit")) as conn:
-            df = pd.read_sql("select * from results", conn)
-            databases.append(df)
+    if isfile(project._transit_database_path):
+        with project.transit_connection as conn:
+            # Check if 'transit_database' has a results table before reading it
+            tables = [x[0] for x in conn.execute("SELECT name FROM sqlite_master WHERE type ='table'").fetchall()]
```
Jake-Moss: As an alterative query 

```python
if conn.execute("SELECT 1 FROM sqlite_master WHERE type ='table' AND name='results'").fetchone():
    ...
```

If the table exists `.fetchone()` will return `(1,)`, otherwise if there is no data (no results table), it will return `None`

Jake-Moss (dismissed):
This is good work. Everything I tried seemed to work pretty well but was I was able get nonsensical results by
- opening the route choice menu, entering the details not but hitting run,
- changing scenario,
- then running the route choice.

The route choice procedure seemed to run on the scenario initially selected, but the results were displayed on the alternative scenario.

I could similarly swap the scenario while performing a transit assignment.

This seems like something pretty tough to solve entirely and globally so maybe it's best to prevent scenario changing when some of the dialogue menus are open? Or to close them when changing, perhaps with a warning? Could this be better solved in AequilibraE?

r-akemii:
This is not intended to happen. My first guess would be "deactivating" the scenario menu selection while other dialogs are open to ensure that we cannot change scenarios while planning to do any action. I'm not sure if this is something to handle with AequilibraE. If deactivating the scenario selection fails, we can think about other options.

qaequilibrae/modules/common_tools/load_graph_layer_setting_dialog.py
```diff
@@ -1,18 +1,16 @@
...
 
 
 class LoadGraphLayerSettingDialog(QtWidgets.QDialog, FORM_CLASS):
-    def __init__(self, iface, project: Project):
+    def __init__(self, qgis_project):
         QtWidgets.QDialog.__init__(self, None, QtCore.Qt.WindowStaysOnTopHint)
-        self.iface = iface
-        self.project = project
+        self.qgis_project = qgis_project
+        self.qgis_project.block_change_scenario()
```
Jake-Moss: There seems to be a couple of `block_change_scenario` calls without a near by `allow_change_scenario`. This is intended?

qaequilibrae/modules/gis/set_color_ramps_dialog.py
```diff
@@ -1,21 +1,24 @@
...
+FORM_CLASS, _ = uic.loadUiType(join(dirname(__file__), "forms/ui_bandwidth_color_ramps.ui"))
 
 
 class LoadColorRampSelector(QDialog, FORM_CLASS):
-    def __init__(self, iface, layer):
+    def __init__(self, qgis_project, layer):
         QDialog.__init__(self)
-        self.iface = iface
+        self.qgis_project = qgis_project
+        self.qgis_project.block_change_scenario()
```
Jake-Moss: Same here

qaequilibrae/modules/transit_procedures/gtfs_feed.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, qgis_project, pt_object):
         self.iface = qgis_project.iface
         self.setupUi(self)
         self.qgis_project = qgis_project
+        self.qgis_project.block_change_scenario()
```
Jake-Moss: And here

Jake-Moss:
That seems to work pretty well, but if an exception is raised between the `block` and `allow` call which prevents `allow` from executing it's possible to lock the scenario change menu. Undoing this requires a full restart of QGIS.

A quick-ish fix might be to try and set the callback for `allow_change_scenario` before, or immediately after the `block_change_scenario` call. So there is never a time when the scenario menu is blocked, but there isn't a way to unblock it. Also I'd make sure the callback is always called. We don't want there to be some way to close the menu without it triggering.

If that doesn't work we could also reach for a mutex or lock object. Typically there used in parallel code sections where only a single thread should have access to something at once, but they're very overkill for this.

qaequilibrae/modules/common_tools/load_graph_layer_setting_dialog.py
r-akemii: Yeah. All these dialogs you mentioned are called inside a parent dialog. Probably blocking scenario changes at these child dialogs is redundant, but the `allow_change_scenario` was removed, so when we go back to the parent dialog, the scenario change menu isn't inadvertently activated. I'll check these later on.

r-akemii:
Good point. I think that adding the callback for `allow_change_scenario` immediately after the `block_change_scenario` would be a nice try. IMO, it would be the easiest way to handle this issue.

r-akemii:
@Jake-Moss I removed the `block_change_scenarios` from all child dialogs, and added a `try/exepct` error handling in all parent dialogs if we have any error between scenario deactivation/activation. Code has another indentation level, but it was the easiest way I found to handle it. If you have a better solution, let me know!

Jake-Moss:
I'm never a huge fan of swallowing all exceptions but I don't really see another way either (it can swallow exceptions that were intended for something else). Ideally I think the exception should be re-raised i.e.

```python
try:
    ...
except Exception as e:
    ...
    raise e
```

This way anything that was intended for code above us still gets the exception it was(n't) excepting, and we can still re-enable the scenario menu. It would also get QGIS handle the exception display instead of us logging it.

For comparison, here is an exception as QGIS logs it, 
image
(errors were produced by attempting to run route choice between two nodes where the origin doesn't exit, just an example of QGIS logging the errors)

and then was as this code logs it while executing the init method. We've lost the traceback, which I always find very useful. (The latest code correctly unblocked the scenario menu this this happened so it's working!)
image
(this error occurs when trying to open the Nauru Visualise data menu, I'll push a commit to fix it)

We also shouldn't return from the `__init__` method once we've handled the exception as this implies that `__init__` succeeded, but in reality the object is now in a bad state when some bits of  `__init__` have run and others haven't. The alternative being raising some exception. I think this might have been fine in this case as the `close` method was being run but I'm not sure.

I'll see if I can make the changes and push directly to this branch.

Jake-Moss:
I think we'll also need to block the scenario menu when all dialogues are open, I've found another way to change break it. I've sent more details over teams.

qaequilibrae/qaequilibrae.py
```diff
@@ -393,11 +426,89 @@ def save_in_project(self):
...
+        """
+        self.cob_scenarios.setEnabled(True)
+
+    def block_change_scenario(self):
+        """
+        Disables scenario changes when QDialogs are open.
+
+        This function is used when initializating classes with dialogs.
+        """
+        self.cob_scenarios.setEnabled(False)
```
Jake-Moss, line 518: Found another fun one, if you
- Open the visualise data,
- load matrix,
- close the matrix dialogue

Then the scenario menu will become unblocked while a dialogue (visualise data) is still open.

To fix this we should use a counter, which is incremented for each block, then decremented for each unblock. This is similar to a semaphore. We only want to unblock the menu when the counter is 0.

I'll push a commit that implements this

pedrocamargo (approved):

PR #481, opened 2025-07-21
Improve installation error messages

### LINKED ISSUE #480: More robust package installation failure messages
When operating behind a firewall, the pip dependency installation may fail and result in an import error when next starting QGIS. This isn't a particularly friendly failure state. I think we should instead check the exit code of the uv/pip subprocess and report a installation failure when it happens, and provide instructions for a manual package installation.

Here is a small script that replicates the package installation method we could provide

```python
# This script is to be executed in the QGIS Python console. Open it via the menu bar ->
# Plugins -> Python console, or use Ctrl + Alt + P. Then use the "Show Editor" button,
# paste this script, edit, and press execute. Pasting into the console directly will
# result in an exception.

import os
import qaequilibrae

# Set the environment variables required for pip here
# os.environ["variable name"] = "variable value"

# Create an instance of the our downloader class, this handles all the dependencies that
# qaequilibrae needs and removes packages that conflict with the global QGIS packages.
installer = qaequilibrae.download_extra_packages_class.DownloadAll()

# The downloader uses two files to mark when dependencies have need installed, we'll
# remove those to get it to retry the downloads.
for file in installer.dependency_files:
    (installer.target_folder / file.name).unlink(missing_ok=True)

installer.install()
```

For the sake of seo: 
qgis ModuleNotFoundError: No module named 'aequilibre'
qgis qaequilibre install failed

r-akemii:
In this PR we change the download extra packages class to handle installation errors, rather than only printing the installation log. We also changed the message when raising an error, and added a FAQ page to QAequilibraE docs to show the script.

### DIFF
--- a/docs/source/faqs.rst
+++ b/docs/source/faqs.rst
@@ -27,9 +27,7 @@ Installation
 
         # The downloader uses two files to mark when dependencies have need installed, we'll
         # remove those to get it to retry the downloads.
-        installer.remove_packages()
-
-        installer.install()
+        installer.retry_pkg_install()
 
 
 User issues
--- a/docs/source/faqs.rst
+++ b/docs/source/faqs.rst
@@ -1,0 +1,44 @@
+Frequently Asked Questions (FAQs)
+=================================
+
+Installation
+------------
+.. admonition:: The installation fails when using a firewall
+
+    When operating behind a firewall, the pip dependency installation may fail and result in an
+    import error when next starting QGIS. One way to ensure the installation, is using the
+    ``DownloadAll`` class in a script, run in the Python terminal.
+
+    .. code::
+
+        # This script is to be executed in the QGIS Python console. Open it via the menu bar ->
+        # Plugins -> Python console, or use Ctrl + Alt + P. Then use the "Show Editor" button,
+        # paste this script, edit, and press execute. Pasting into the console directly will
+        # result in an exception.
+        import os
+        from qaequilibrae.download_extra_packages_class import DownloadAll
+
+        # Set the environment variables required for pip here
+        # os.environ["variable name"] = "variable value"
+
+        # Create an instance of the our downloader class, this handles all the dependencies that
+        # qaequilibrae needs and removes packages that conflict with the global QGIS packages.
+        installer = DownloadAll()
+
+        # The downloader uses two files to mark when dependencies have need installed, we'll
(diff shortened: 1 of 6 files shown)


docs/source/faqs.rst
```diff
@@ -0,0 +1,44 @@
...
+        # Set the environment variables required for pip here
+        # os.environ["variable name"] = "variable value"
+
+        # Create an instance of the our downloader class, this handles all the dependencies that
+        # qaequilibrae needs and removes packages that conflict with the global QGIS packages.
+        installer = DownloadAll()
+
+        # The downloader uses two files to mark when dependencies have need installed, we'll
+        # remove those to get it to retry the downloads.
+        installer.remove_packages()
```
pedrocamargo:
```suggestion
        installer.retry_pkg_install()
```

docs/source/faqs.rst
```diff
@@ -0,0 +1,44 @@
...
+
```
pedrocamargo:
```suggestion

```

docs/source/faqs.rst
```diff
@@ -0,0 +1,44 @@
...
+        installer.install()
```
pedrocamargo:
```suggestion
```

qaequilibrae/download_extra_packages_class.py
```diff
@@ -158,11 +163,18 @@ def clean_packages(self, target_folder):
                             f"Duplicated packages removed from installation: {fldr}", level=Qgis.MessageLevel.Info
                         )
 
+    def remove_packages(self):
```
pedrocamargo:
```suggestion
    def retry_pkg_install(self):
```

qaequilibrae/download_extra_packages_class.py
```diff
@@ -158,11 +163,18 @@ def clean_packages(self, target_folder):
...
+        existing_files = list(os.walk(self.target_folder))[0]
+        for packages in existing_files[1]:
+            shutil.rmtree(self.target_folder / packages)
+
+        for file in existing_files[2]:
+            if file == "__init__.py":
+                continue
+            (self.target_folder / file).unlink()
```
pedrocamargo, line 174:
```suggestion
            (self.target_folder / file).unlink()
        self.install()
```

pedrocamargo (approved):

PR #500, opened 2025-10-07
Export traffic assignment configurations

### LINKED ISSUE #251: [New Feature] Implement YAML export during assignment
Implement a YAML export during assignment and, then, run assignment using QGIS processing provider with a function just like https://github.com/AequilibraE/qaequilibrae/pull/250 (would take the YAML as only input).

r-akemii:
Closes #251.


pedrocamargo:
@Jake-Moss , can you take a quick look on this one?

qaequilibrae/modules/paths_procedures/create_py_strings.py
```diff
@@ -0,0 +1,78 @@
...
+    sl = dct.get("select_links")
+    if sl:
+        select_link = """\n\tfor tc in traffic_classes.values():\n\t\ttc.set_select_links({})"""
+        func_string += select_link.format(sl["select_links"])
+
+    # Execute procedure
+    func_string += "\n\tassig.execute()\n"
+
+    # Save outputs
+    func_string += f"\n\tassig.save_results('{dct.get("scenario_name")}')"
```
Jake-Moss: Mixing quotes within an f-string is only supported on 3.12+. I'm not sure what the Python versions of the QGIS versions we support are

qaequilibrae/modules/paths_procedures/create_py_strings.py
```diff
@@ -0,0 +1,78 @@
...
+        lines = file.readlines()
+
+    # Find the last import statement to insert after
+    import_line_idx = -1
+    for idx, line in enumerate(lines):
+        if line.strip().startswith("import") or line.strip().startswith("from"):
+            import_line_idx = idx
+    insert_idx = import_line_idx + 1
+
+    lines.insert(insert_idx, f"from .{Path(out_name).stem} import run_assignment\n")
```
Jake-Moss: This will break with multi-line imports, e.g.

```python
from foo import (
    bar,
    baz,
)
```

Jake-Moss:
> @Jake-Moss , can you take a quick look on this one?

Missed that first review request, sorry.

I like the export to yaml functionality, but I'm not sold on the "create a python file" method. String formatting to generated code is incredibly fragile. A single `'` in any of the fields would break the file entirely. Escaping the strings properly is a tedious process as well.

As it stands the strings used to template the code are difficult to read and reason about. I think we'd be better served by a templating library if this was the path we wanted to go. Jinja2 and Tempita (used by Cython) are popular.

However, what does the "generate a Python function" method offer that the yaml export doesn't? Is it intended to be edited by users? I.e. generated once -> modify yourself? If so there’s no check to make sure the file doesn't override something that already exists. It's also possible that the users selects the `__init__.py` file accidentality, then creates a malformed file. 

If the intent is generate file -> use cli to run it via https://github.com/AequilibraE/aequilibrae/pull/726. I think we'd be much better served by providing a method to load an run one of the yaml exports.

The generated files are also specific to the version of [q]aequilibrae that was used to generate them. If we change the API in an even slightly breaking way we have no way to provide backwards compatibility to them. The user would have to fix the file themselves. With the yaml export, the importer could handle a lot of the changes without concerning the user.

I think it's much more useful to offer an example file in the documentation that can be used as a starting place for an assignment procedure or provide a way to execute an exported configuration via the run module.

@pedrocamargo are method configuration files like the yaml export something we want to do more? Like a file provides a set of parameters and a generic procedure reads and executes it?

qaequilibrae/modules/paths_procedures/create_py_strings.py
r-akemii: Thanks for warning me, Jake! I just checked and the Python version for QGIS 3.34.10 (which is out minimum version) is 3.12.5. I'll change this string (even for readability), but in theory it should work in all versions from 3.34.10.

pedrocamargo:
@Art-Ev , this would probably interest you

pedrocamargo (approved):

Art-Ev:
that's very cool ! 😍

PR #505, opened 2025-10-16
Change dependencies installation folder

r-akemii:
The rationale behind this PR is avoid the installation of external dependencies on the QAequilibraE's package folder, separating QAequilibraE's dependencies from project specific dependencies.


r-akemii:
@pedrocamargo I removed the class logger because every use of 'Run procedures' generated one or more inputs in the logger that were not so helpful at all. I replaced it for printing the results' output when available with pprint in the QGIS message tab and showing this information to the user. If you think it's necessary return the logger, I can fix it.

pedrocamargo (approved):

PR #509, opened 2025-10-21
Fix paper header

r-akemii:
In this PR we add headers for the paper.


r-akemii:
@pedrocamargo can we add the headers once again because it contains all paper metadata and reference configuration for submission? I'll remove the 'Draft PDF' worklflow when we're able to aprove the paper :)

pedrocamargo (approved):

PR #517, opened 2025-11-12
Fix error when cancelling file selection dialog 

### LINKED ISSUE #516: [BUG] GetOutputFileName returns an AttributeError when closing the window
**Describe the bug**
If the GetOutputFileName window is open and we press "ESC", an error is raised.

**To Reproduce**
1. Open a project
2. Go to Transit -> Import GTFS
3. Click on "Add feed"
4. When the *Target GTFS feed* window opens, press ESC.
5. The error message shows up

**Expected behavior**
I would expect to see no error message. Maybe adding a message at the log message tab could be useful.
Other issue (at least when this error is raised by the GTFS) is that the scenario change remains blocked,

**Screenshots**
Image

 - QGIS Version 3.40.10

**Other notes**
- Check if this behaviour happens every time we use `GetOutputFileName`
- Adding a test for the function is a plus

tomstephen:
In this PR we fix the error that occurs when cancelling the selection of a file within the dialog box, e.g., when selecting the GTFS feed. Updated get_output_file_name.py to handle cancelled dialog, added test.


qaequilibrae/modules/common_tools/get_output_file_name.py
```diff
@@ -22,7 +23,10 @@ def GetOutputFileName(clss, box_name, file_types, default_type, start_path):
             extension = new_name[-3:]
         else:
             extension = new_name[-4:]
-    return new_name, extension.upper()
+        return new_name, extension.upper()
+    else:
+        QgsMessageLog.logMessage("No file selected.", "Messages", level=Qgis.MessageLevel.Info)
```
r-akemii: This message is not as useful as I first thought. We can remove it and the unused imports.

test/test_pt_procedures.py
```diff
@@ -41,6 +43,26 @@ def test_add_new_feed(pt_no_feed, mocker):
     assert var == 1
 
 
+def test_get_output_file_name_no_selection(mocker):
```
r-akemii: Let's create a test file for `GetOutputFileName` and test both cases: with and without file selection.

r-akemii:
We still have to make a few changes to make this a great PR

- `GetOutputFileName` is also raising the same error in other classes like traffic assignment, distribution models (we can check them later)
- When we exit the procedure in GTFS feed, its dialog remains open. We need to add a fix for this too.

test/test_common_tools.py
```diff
@@ -0,0 +1,57 @@
...
+
+#     assert name is None
+#     assert ext is None
+
+
+@pytest.mark.parametrize(
+    ("is_filename_chosen", "filename", "exp_fname", "exp_ext"),
+    [
+        (True, "target.txt", "target.txt", "TXT"),
+        (True, "document..md", "document.md", "MD"),
```
r-akemii, line 29: Quick question: the two dots in the file name are intentional or a typo?

pedrocamargo (approved):

PR #540, opened 2026-08-11
Fix crash opening a QGIS project saved with a scenario active

pedrocamargo:
_project_base_path tested for '/scenarios/' in a str(Path), which never matches on Windows, so the active scenario folder was stored as aequilibrae_path. Reopening then tried to open a scenario database as a project and failed on the missing scenarios table. Use root_scenario.base_path instead, map already-saved scenario paths back to the project root on load, and skip a stale scenario rather than aborting the QGIS project open.


pedrocamargo (commented):
.

PR #577, opened 2026-08-18
Delete matrix and result records from the data viewer

pedrocamargo:
Double-clicking a row in either the Matrices or the Results tab of the Visualize data screen now deletes that record, after a confirmation box that defaults to No so a stray double-click or Enter cannot destroy anything.

Both helpers delegate to AequilibraE, which owns what deleting means: a matrix record goes along with its file on disk, and a result record along with its table in the results database. Two cases it cannot cover are handled here. Its gateways are built when the project is opened, so they are reloaded first, otherwise anything produced during the session looks like a record that does not exist. Matrix records whose file is missing from disk never enter that gateway at all - exactly the rows the viewer flags as not found on disk - so those orphans are removed straight from the project database.

Results recorded in the transit database are outside the project's Results gateway but are listed by the viewer, so they are removed here with the same semantics.

list_results concatenated the project and transit results without resetting the index, leaving duplicate labels. The table view addresses rows by position, so with transit results present a label lookup reached the wrong record, which would have deleted the wrong result. The same mismatch was already reachable through the Load Result table button.

### DIFF
--- a/.github/workflows/AequilibraeUpdates.yml
+++ b/.github/workflows/AequilibraeUpdates.yml
@@ -37,4 +37,4 @@ jobs:
           echo "PYTHONPATH=$PYTHONPATH" >> $GITHUB_ENV
           export QT_QPA_PLATFORM=offscreen
           pwd
-          python3 -m pytest --cov-report term --cov-config=.coveragerc --cov=qaequilibrae test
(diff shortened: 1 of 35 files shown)


tomstephen:
Now passing. Once again, failing action is not caused by failing tests. Needs #586 merged first.

PR #578, opened 2026-08-18
Minor bugs

pedrocamargo:
1. A project created from scratch showed no layers in the panel

Every project-creation path handed the new project to the panel except the Processing algorithm Create empty project, which created the folder and closed the project immediately. The panel kept showing whatever it had before, so the new model's layers were nowhere to be found.

2. Two traffic classes reading matrix cores of the same name killed the assignment

Result columns are named after the matrix core, not the traffic class. Two classes reading cores of the same name — two matrices both carrying a core called matrix is routine — produced duplicate columns, and saving died with duplicate column name: matrix_ab. It failed at save time, with the whole assignment already run, inside a Qt slot with no handler. The "class name already used" guard was also missing its return, so class names were never actually unique.

3. Scenario comparison crashed on every field except the flow columns

AttributeError: 'NoneType' object has no attribute 'real'. The field list was lower-cased for the dropdown while the result table keeps its real column names, and AequilibraE capitalises everything it computes itself (Preload_AB, VOC_AB, Congested_Time_AB). The case-sensitive match never found them, so no data field reached the layer, QgsExpression.evaluate() returned NULL, and reading .real off it blew up. Only the flow columns — lower-case, named after matrix cores — ever worked.

4. Class names differing only by case hit the same wall as #2 (from review)

SQLite rejects two columns that differ only by case, so classes named car and Car passed every guard and then failed at save time — the same failure #2 set out to remove. Whitespace-only names also slipped through the emptiness check.

### DIFF
--- a/qaequilibrae/modules/common_tools/live_log.py
+++ b/qaequilibrae/modules/common_tools/live_log.py
@@ -62,7 +62,11 @@ class LiveLogWidget(QWidget):
             return
         self.text.appendPlainText(msg)
         if self.auto_scroll.isChecked():
-            self.text.ensureCursorVisible()
+            # Drive the scrollbar rather than the cursor. appendPlainText leaves the
+            # text cursor where it was -- usually the very start -- so asking to make
+            # the cursor visible scrolls back to the top instead of following the tail.
+            scrollbar = self.text.verticalScrollBar()
+            scrollbar.setValue(scrollbar.maximum())
 
     def connect_bridge(self, bridge):
         bridge.log_line.connect(self.append)
--- a/qaequilibrae/modules/common_tools/live_log.py
+++ b/qaequilibrae/modules/common_tools/live_log.py
@@ -62,8 +62,7 @@ class LiveLogWidget(QWidget):
             return
         self.text.appendPlainText(msg)
         if self.auto_scroll.isChecked():
-            # Drive the scrollbar rather than the cursor. appendPlainText leaves the
-            # text cursor where it was -- usually the very start -- so asking to make
+            # appendPlainText leaves the text cursor where it was so asking to make
             # the cursor visible scrolls back to the top instead of following the tail.
             scrollbar = self.text.verticalScrollBar()
             scrollbar.setValue(scrollbar.maximum())
--- a/qaequilibrae/modules/common_tools/table_field_lister.py
+++ b/qaequilibrae/modules/common_tools/table_field_lister.py
@@ -2,10 +2,9 @@ from sqlite3 import Connection
 
 
 def find_table_fields(conn: Connection, table_name: str):
-    """Returns the lower-cased column names of *table_name*.
+    """Returns the column names of *table_name*, spelled as the table spells them.
 
     Uses the table-valued form of the pragma so the table name travels as a bound parameter
-    rather than being interpolated into the statement.
-    """
+    rather than being interpolated into the statement."""
     structure = conn.execute("SELECT name FROM pragma_table_info(?);", [table_name]).fetchall()
-    return [x[0].lower() for x in structure]
+    return [x[0] for x in structure]
--- a/qaequilibrae/modules/common_tools/table_field_lister.py
+++ b/qaequilibrae/modules/common_tools/table_field_lister.py
@@ -2,7 +2,7 @@ from sqlite3 import Connection
 
 
 def find_table_fields(conn: Connection, table_name: str):
-    """Returns the column names of *table_name*, spelled as the table spells them.
+    """Returns the column names of *table_name*.
 
     Uses the table-valued form of the pragma so the table name travels as a bound parameter
     rather than being interpolated into the statement."""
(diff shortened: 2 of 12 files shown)


qaequilibrae/modules/common_tools/table_field_lister.py
```diff
@@ -2,10 +2,9 @@
 
 
 def find_table_fields(conn: Connection, table_name: str):
-    """Returns the lower-cased column names of *table_name*.
+    """Returns the column names of *table_name*, spelled as the table spells them.
```
tomstephen:
```suggestion
    """Returns the column names of *table_name*.
```

I would expect than returning them as-is to be the default, so we don't need the extra text.

qaequilibrae/modules/gis/compare_scenarios_dialog.py
```diff
@@ -15,6 +15,18 @@
...
 
 
+def directional_field_pairs(fields):
+    """AB/BA pairs as (label, (ab, ba)), matched case-insensitively and named as the table names them."""
+    by_lower = {f.lower(): f for f in fields}
+    pairs = []
+    # Suffix rather than substring, so a class called "cab" is not paired with itself
+    for lower, name in by_lower.items():
+        counterpart = by_lower.get(f"{lower[:-2]}ba") if lower.endswith("ab") else None
+        if counterpart is not None:
```
tomstephen:
```suggestion
        if counterpart:
```

qaequilibrae/modules/gis/compare_scenarios_dialog.py
```diff
@@ -196,6 +208,24 @@ def execute_comparison(self):
...
+        value = exp.evaluate(self.link_layer.createExpressionContext())
+        if value is not None:
+            return float(value)
+
+        reason = exp.evalErrorString() or exp.parserErrorString() or self.tr("no value came back")
+        self.qgis_project.iface_error_message(
+            self.tr("Could not measure the fields being compared: {}").format(reason),
+            self.tr("Scenario comparison"),
+        )
+        # Both are expression text by now, and a second run reads them as numbers again
```
tomstephen:
```suggestion
        # Reset `band_size` and `space_size`
```
Vestigial comment? We should explain what the next few lines do, as its not obvious by the names of the methods.

qaequilibrae/modules/gis/compare_scenarios_dialog.py
```diff
@@ -226,26 +256,25 @@ def load_result_tables(self):
         v2 = self.cob_alt_scenario.currentText()
         v3 = self.cob_base_result.currentText()
         v4 = self.cob_alternative_result.currentText()
-        v5 = self.cob_base_data.currentText()
-        v6 = self.cob_alternative_data.currentText()
+        # The real column names, not the starred labels
```
tomstephen:
```suggestion

```

qaequilibrae/modules/menu_actions/load_project_action.py
```diff
@@ -44,6 +44,7 @@ def _run_load_project_from_path(qgis_project, proj_path):
         else:
             raise e
 
+    qgis_project.project = project
     show_project_in_panel(qgis_project, proj_path)
```
tomstephen: If we open the project here, then something in `show_project_in_panel` fails (not unrealistic, as we're dealing with an Ad-hoc project upgrade...), the project will get stuck as open but empty.

qaequilibrae/modules/paths_procedures/traffic_assignment_dialog.py
```diff
@@ -260,7 +262,7 @@ def export_yaml(self):
                 pth = Path(info.matrix.file_path).name
                 df = self.project.matrices.list()
                 dc["matrix_name"] = df.loc[df["file_name"] == pth]["name"].values[0]
-                dc["matrix_core"] = info.matrix.view_names[0]
+                dc["matrix_core"] = self.class_cores[tc][0]
```
tomstephen: Takes the first core per class - can a class have multiple cores? If one does, they get silently dropped

qaequilibrae/modules/menu_actions/load_project_action.py
tomstephen: I think we need to handle the exceptions of both parts together, so it either all works or all fails

qaequilibrae/modules/common_tools/live_log.py
```diff
@@ -62,7 +62,11 @@ def append(self, msg):
             return
         self.text.appendPlainText(msg)
         if self.auto_scroll.isChecked():
-            self.text.ensureCursorVisible()
+            # Drive the scrollbar rather than the cursor. appendPlainText leaves the
+            # text cursor where it was -- usually the very start -- so asking to make
+            # the cursor visible scrolls back to the top instead of following the tail.
+            scrollbar = self.text.verticalScrollBar()
```
tomstephen:
```suggestion
            # appendPlainText leaves the text cursor where it was so asking to
            # make the cursor visible scrolls back to the top instead of
            # following the tail.
            scrollbar = self.text.verticalScrollBar()
```

Just need to comment the current state of things, not the change

tomstephen (commented):
Changes work nicely, I'll tidy up these few comments in the morning

PR #579, opened 2026-08-18
Add link types and modes

pedrocamargo:
Two GUIs under the Data group of the Processing toolbox, for adding a link type and a mode to the open project. Neither gets a menu entry, so the toolbox is the only route to them.

Both dialogs list what the project already has, offer the first single-letter identifier still free, and check the identifier and the name before handing them to AequilibraE - the database only enforces uniqueness and the length of the identifier, and AequilibraE only looks at the name once the record is already being written.

The listing follows the project's own table rather than AequilibraE's schema, because the two disagree: Sioux Falls and create_example() carry no link_types.speed but do carry alpha through tau. Writing to a column the table does not have is silently dropped, so the form hides the inputs this project cannot store. After each addition the links layer's attribute form is rebuilt, so a new mode or link type can be picked straight away when digitizing.


qaequilibrae/modules/processing_provider/data_procedures/add_mode.py
```diff
@@ -0,0 +1,26 @@
...
+            "data",
+            self.tr("Adds a mode to the network of the open project"),
+            ["mode", "modes", "network", "add", "new"],
+        )
+
+    def createInstance(self):
+        return AddMode()
+
+    def tr(self, message):
+        return trlt("AddMode", message)
```
tomstephen, line 31:
```suggestion
class AddMode(QAequilibraEProcessingAlgorithm):
    def __init__(self):
        from qaequilibrae.modules.menu_actions import run_add_mode

        super().__init__(
            partial(run_add_mode, get_aequilibrae_menu_instance()),
            "add_mode",
            self.tr("Add mode"),
            self.tr("Data"),
            "data",
            self.tr("Adds a mode to the network of the open project"),
            ["mode", "modes", "network", "add", "new"],
        )
        
    def flags(self):
        return super().flags() | Qgis.ProcessingAlgorithmFlag.NoThreading

    def createInstance(self):
        return AddMode()

    def tr(self, message):
        return trlt("AddMode", message)
```

As per [these docs](https://docs.qgis.org/3.16/en/docs/pyqgis_developer_cookbook/tasks.html), we need to keep processing algorithms on the same thread. Same comment for `add_link_type.py`

qaequilibrae/modules/network/add_mode_dialog.py
```diff
@@ -0,0 +1,52 @@
...
+        mode = modes.new(mode_id)
+        mode.mode_name = name
+        mode.description = self.txt_description.text().strip() or None
+        mode.pce = self.dsb_pce.value()
+        mode.vot = self.dsb_vot.value()
+        mode.ppv = self.dsb_ppv.value()
+
+        try:
+            modes.add(mode)
+            mode.save()
```
tomstephen, line 44: If the `.add` passes and the `.save` fails, we get a partially instantiated mode, and can't try again to overwrite it. I think I have a fix, next commit

qaequilibrae/modules/network/add_network_record_dialog.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def report(self, message: str, is_error: bool = False):
+        self.lbl_feedback.setText(message)
+        self.lbl_feedback.setStyleSheet("color: red;" if is_error else "")
+
+    def refresh_link_editing_form(self):
+        """The links layer offers the modes and link types read when it was added to the canvas."""
+        links = self.qgis_project.layers.get("links")
+        if links is not None:
+            load_editor_styles(links[0], "links", self.project)
```
tomstephen, line 104: If we have `n` modes, there are (I believe...) 2^n - 1 unique combinations, so for 10 modes you would get 1023 entries, etc. Do we want to limit the number of modes to stop it from getting unreasonable here? Or is it reasonable to assume you'll only have a few modes?

EvelynBond (approved):

PR #582, opened 2026-08-24
Fix package installation on MacOS

tomstephen:
This PR,

1. Fixes where we look for a Python executable - the path for "Darwin" (macOS) is different to Linux/Windows.
2. Builds the AequlibraE package separately, then copies across. AequilibraE does not provide wheels for macOS.
    1. This requires some manual setup by the user (installing brew, uv, llvm, spatialite), and hence *docs are added to reflect this*.
    2. Building wheels for AequilibraE was discussed by decided to be too much effort for now, as building portable wheels reliant on brew is fragile, and so a new build system would be needed.

Tested on my M5 Macbook (26, Tahoe) and M1 Mac Mini (15, Sequoia), with QGIS 3.44 and 4.2. Untested on Intel macOS, but the paths for homebrew come from the [homebrew docs](https://docs.brew.sh/Installation) and check both Apple Silicon and Intel paths, so expect it to work. If this becomes an issue in the future for someone, I have a 2015 Macbook I can dig out...

### DIFF
--- a/qaequilibrae/download_extra_packages_class.py
+++ b/qaequilibrae/download_extra_packages_class.py
@@ -40,9 +40,10 @@ class DownloadAll:
         self.error = 0
 
     def install(self):
-        command = [str(self.find_python()), "-m", "pip", "install", "uv"]
-        _ = self.execute(command)
-        print(" ".join(command))
+        if sys.platform != "darwin":
+            command = [str(self.find_python()), "-m", "pip", "install", "uv"]
+            _ = self.execute(command)
+            print(" ".join(command))
 
         for file in self.dependency_files:
             flag = self.target_folder / file.name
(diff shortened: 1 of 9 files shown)


qaequilibrae/requirements.txt
```diff
@@ -1 +1,2 @@
-ortools
\ No newline at end of file
+ortools
```
pedrocamargo: I don't even know if we still use ortools!!!  We used to have a tool to create vehicle routes using it, back in the day!

tomstephen:
Unrelated test failure, its the overpass API rate limit

qaequilibrae/download_extra_packages_class.py
```diff
@@ -38,11 +39,15 @@ def __init__(self):
         self.target_folder = pth / "packages"
         self.no_ssl = False
         self.error = 0
+        self.last_exit_code = 0
 
     def install(self):
-        command = [str(self.find_python()), "-m", "pip", "install", "uv"]
-        _ = self.execute(command)
-        print(" ".join(command))
+        self.target_folder.mkdir(parents=True, exist_ok=True)
```
EvelynBond, line 45: self.target_folder is already getting a director made- see line 66/79

EvelynBond (approved):
Overall seems good. 

It looks like you've repeated making a directory- I've left a comment about that.

pedrocamargo (approved):

qaequilibrae/requirements.txt
tomstephen: Okay excellent, will remove. Unrelated to this PR (I think this comment might have come with a linting mistake was checked in?), I will add to the list

PR #586, opened 2026-09-08
Remove qtbot and waits

tomstephen:
In an attempt to fix some potential race conditions (tests failing in CI on #577, working elsewhere and locally, unrelated changes) I've replaced the usage of `qtbot` and waits with ~mocking~ event filter and queueing for more deterministic CI behavior. 

~I have a solution for keeping the whole `.exec` workflow, which then wouldn't require mocking, but it's a more major change and I want to see if this fixes it first.~


tomstephen:
Unrelated test failure, the tests actually pass

PR #597, opened 2026-09-16
Add shortest path processing algorithm

tomstephen:
Demo on how we may abstract some functionality to processing algorithms.


tomstephen:
We get a really nice double-win for this work, both in making a clearer delineation between GUI and geoprocessing/aequilibrae, and opening up more "QGIS-esque" workflows for users (e.g., model designer) which in turn makes QAequilibraE and AequilibraE more accessible to more people.

Given the scope to tackle this properly, I'm thinking a separate branch for `processing_algorithms` then we (fyi @EvelynBond) can progressively refactor the code base by adding to that. 

I'll finish the remaining issues then open that tract of work.

pedrocamargo:
Looks good, @tomstephen

tomstephen:
This work is happening on #602

Issue #595, opened 2026-09-15
Collate logs into a unique category


tomstephen:
We under-use logging currently, making it harder to understand/track what's happening behind the scenes.

It is possible to add a new category to the existing Log Messages tabs ("Plugins", "General", etc) like so

```python
from qgis.core import QgsMessageLog, Qgis

QgsMessageLog.logMessage("Here is my useful message", "QAequilibraE", Qgis.Info)
QgsMessageLog.logMessage("Another useful message", "QAequilibraE", Qgis.Info)
```
[ref](https://qgis.org/pyqgis/3.44/core/QgsMessageLog.html#qgis.core.QgsMessageLog.logMessage)

Suggest we add our own logging method that accepts message and level, then calls `QgsMessageLog.logMessage` as above.

tomstephen:
Closed by #598

PR #598, opened 2026-09-18
ty, menubar, logging

tomstephen:
This PR,

- enables `ty` (disabled outside of processing_provider folder, for now). Run locally with `just lint`, runs in CI
- uses Python standard loggers, connects a handler to the QGIS Message Log setup, with an `AequilibraE` category. Little use currently, but in place for adding more comprehensive logs (ie., in following refactor...)
- removes all processing-algorithms-that-aren't-really-processing-algorithms
- adds an `AequlibraE` menubar to be analogous to the quick-access commands people are familiar with from the (misused) processing toolbox`

All to prepare for processing algorithm refactor.


tomstephen:
Closes #595

PR #603, opened 2026-09-21
algorithm for opening (and hopefully closing) an aequilibrae project

EvelynBond:


tomstephen:
Nice work on this one, had a try and quite like it. 

One issue however, say we want to open a project, run two different traffic assignments, do something with those ouputs, then close the project - how do we ensure the project is not closed until *after* the assignments (and anything else reliant on the project) is run? To QGIS, it just looks like another processing algorithm, so unless we hand-set the [dependencies](https://docs.qgis.org/3.44/en/docs/user_manual/processing/modeler.html#definition-of-the-workflow), it could realistically close the project before it runs the other algorithms. We would also give up the nice multiprocessing support we would otherwise have.

Let's park this for now, and use paths to AequilibraE projects, opening & closing per algorithm.

